PR #824, opened 2014-12-18
Switch from 'minifier' to 'minify' for build minification (fixes #817)

### LINKED ISSUE #817: Syntax error in minified build
Safari throws an error using `remotestorage.min.js`:

     [Error] SyntaxError: Cannot declare a parameter named 'e' as it shadows the name of a strict mode function.

Version 0.11.0
Safari 7.1 (Mac OS 10.9.5) and Safari iOS 8

galfert:
This uses the 'minify' library (http://coderaiser.github.io/minify/), because the builds produced with 'minifier' (https://github.com/fizker/minifier) had a syntax error in Safari.

After merging this, you'll have to delete the `./node_modules` directory before running `npm install`. The command-line script in `./node_modoules/.bin/` has the same name as the previous one.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,7 @@ compile-assets: $(ASSETS_OUT)
 
 %.min.js: %.js
 #	minify $< -o $@ --mangle --wrap --export-all
-	$(MINIFY_BIN) -o $@ $<
+	$(MINIFY_BIN) $< > $@
 	mv $@ $@.tmp
 	head -n1 $< > $@
 	cat $@.tmp >> $@
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   "homepage": "http://remotestorage.io",
   "devDependencies": {
     "jaribu": "0.x.x",
-    "minifier": "^0.6.0"
+    "minify": "^1.4.0"
   },
   "scripts": {
     "test": "node_modules/jaribu/bin/jaribu"


michielbdejong:
> it's actually fine not be able to minify it again

I don't agree that it's fine, although I think if minifying our library breaks it, then it's the fault of the minifier used. If this is the case, then we should file a bug to the minifier in question, and reference that as a known regression ("no longer minifiable with this-and-that minifier due to <link>") in our changelog, and also recommend a minifier that is able to minify our library without problems.

skddc:
Ok, can you take over this issue?

silverbucket:
BTW I followed up with the maintainer of `minifier` here, https://github.com/fizker/minifier/issues/8

Seems he just uses `uglify` internally, but just with some more strict compression settings ( i think it has to do with removing semicolons, seems like that would make it more difficult to minify a second time ).

Anyway, perhaps we should use uglify but use as many options as we can to shrink the size, a normal uglification without special options makes a pretty huge minified file.

skddc:
So we can't just set an option for minifier to not delete semicolons? Seems like a dangerous thing to do anyway (and obviously is, as we see here).

silverbucket:
@skddc well I'm not certain that's the issue, just could be. Probably saves a lot of space though too. it's not difficult to make re-minimizable JS builds, it's just finding the balance between as small as we can get it without breaking re-minification. 

As for `minifier` if it's just a wrapper for uglify, there's no real reason to use it directly if we're going to pick out uglify options to use anyway, may as well just use `uglify` directly.

skddc:
K, thanks for the clarification!

galfert:
Replaced by #832.

PR #830, opened 2015-01-22
Use stateSetter function in widget error handling function (fixes #829)

### LINKED ISSUE #829: ""[Sync] Error" TypeError: widget.view is undefined" when not using widget
Error logged for every sync operation, while sync works in general.

skddc:
When not using the widget, there's no widget.view, so the direct setting of view state in the errorHandling function doesn't work. This makes it use the stateSetter helper function, which actually checks for
widget.view presence before trying to set properties on the view.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ All releases can also be found and downloaded on the
 ### Bugfixes
 
 * Fix single quotes in filenames not working correctly (#812)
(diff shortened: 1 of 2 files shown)


silverbucket:
+1 looks good to me

galfert:
+1

skddc:
Thanks, guys!

PR #831, opened 2015-01-22
Fix bug introduced in recent syntax clean-up

skddc:
This bug was introduced in the recent syntax cleanup session. It changes a line of code that handles objects into one that handles arrays. This is both fixing the line as well as refactoring the original one to be slighly nicer.


galfert:
+1

silverbucket:
+1

skddc:
andibelike

PR #832, opened 2015-01-22
Switch from 'minifier' to 'uglifyjs' for build minification (fixes #817)

### LINKED ISSUE #817: Syntax error in minified build

galfert:
The filesize of the minified build is bigger than before, but at least there are no syntax errors. We might be able to make it smaller using some other options later.

This replaces the PR #824, the minified builds can still be run through other minifiers without breaking.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ DOC_BIN        = naturaldocs
 DOC_DIR        = ./doc/code
 DOC_CONFIG_DIR = ./doc/config
 DOC_CUSTOM_CSS = custom-1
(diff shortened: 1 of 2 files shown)


skddc:
+1

silverbucket:
+1

skddc:
dancing-ape

Issue #740, opened 2014-08-06
Offline mode broken


xMartin:
When I go offline the widget is not indicating offline state.

Broken in 0.10.0, works in 0.10.0-beta.

Tested with Work Offline in Firefox. Apps: Grouptabs, Sharesome (0.10.0) and Litewrite (0.10.0-beta).

skddc:
This is (still?) broken in the latest stable release in FF for me. I have a feeling this never worked since Martin reported it.

skddc:
Can someone please help verify this issue?

skddc:
Ok, I tested in both FF and Chrome and the offline mode gets set correctly (`remoteStorage.remote.online` is set to `false` after failing requests and `true` again after the next successful one). However the widget offline state isn't set.

skddc:
https://github.com/remotestorage/remotestorage.js/blob/master/src/widget.js#L194 is hit as well.

xMartin:
Works for me in 0.11.2. Cool!

skddc:
\o/

PR #840, opened 2015-02-13
Fix widget not being set to correct modes in errorsHandler (refs #740)

skddc:
stateSetter actually returns a function instead of setting the state immediately, so it needs to be applied from errorsHandler.

You can try this by turning off your Internet connection and seeing that the widget actually goes into offline mode, as well as going back when turning the connection back on.

This is kind of urgent, because it's basically broken in all apps using >=0.11.0 so quick reviews would be helpful!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All releases can also be found and downloaded on the
 ### Bugfixes
 
 * Fix usage of FileReader that doesn't support `addEventListener` (#836)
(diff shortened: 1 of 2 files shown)


skddc:
Hah, dat hound got some work done. :)

All done, though. Please review. Thanks!

skddc:
Can someone please review? Thanks!

michielbdejong:
+1

galfert:
LGTM. +1

skddc:
Thanks.

PR #842, opened 2015-02-16
Don't use incompatible/buggy IndexedDB implementations (fixes #841)

### LINKED ISSUE #841: Problems with incompatible indexedDB implementations
When using remoteStorage in a Cordova app on Android versions before 4.4 (or in the stock Android browser), the feature detection for IndexedDB doesn't work properly. Neither the `onerror` nor the `onsuccess` handlers ([see here](https://github.com/remotestorage/remotestorage.js/blob/0.11.1/src/indexeddb.js#L336-L342)) are called. Because of that the corresponding promise is never fulfilled or rejected, leaving the remoteStorage instance in some kind of limbo state.

galfert:
This PR checks for Browsers with incompatible IndexedDB implementations (namely the stock Browser of all Android versions prior to 4.4).

I couldn't find any specific checks to detect problematic implementations, so I had to look for specific user agents. If anybody knows of better ways to do this, I would be glad to hear them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ All releases can also be found and downloaded on the
 
 * Fix usage of FileReader that doesn't support `addEventListener` (#836)
 * Fix widget not being set to correct mode in some cases (e.g. offline) (#740)
+* Don't use incompatible/buggy IndexedDB implementations (#841)
 
 ## 0.11.1 (January 2015, Hacker Beach Edition)
 
--- a/src/indexeddb.js
+++ b/src/indexeddb.js
@@ -330,7 +330,21 @@
   RS.IndexedDB._rs_supported = function () {
     var pending = Promise.defer();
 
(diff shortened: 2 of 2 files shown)


skddc:
+1

silverbucket:
It would be better if we could test some specific, easy, check against the implementation itself. Though I guess different browsers will have different problems with their implementation. This essentially whitelists only Chrome and Firefox on Android devices. What about Dolphin and Opera(?) and other browsers which might be OK? Might it be better to blacklist the ones we know are bad and leave unknowns alone so users could use them if they work OK?

skddc:
@silverbucket That's the problem. We *do* test opening a database, and that fails without calling any callbacks or throwing any exceptions. The implementation is just completely broken, so we need to fall back in any case for that engine.

skddc:
@silverbucket have a minute to look at this again?

silverbucket:
+1

PR #844, opened 2015-02-16
Use BlobBuilder instead of Blob constructor in older Browsers (fixes #843)

### LINKED ISSUE #843: "Illegal constructor" in older Android Browsers
Seeing this error when trying to create an instance of `Blob` with `new Blob()`.

galfert:
This fix prevents the `Illegal constructor` error for Browsers that don't support the Blob constructor.

### DIFF
--- a/src/wireclient.js
+++ b/src/wireclient.js
@@ -93,10 +93,19 @@
   }
 
   function readBinaryData(content, mimeType, callback) {
-    var blob = new Blob([content], { type: mimeType });
+    var blob;
+    global.BlobBuilder = global.BlobBuilder || global.WebKitBlobBuilder;
+    if (typeof global.BlobBuilder !== 'undefined') {
+      var bb = new global.BlobBuilder();
+      bb.append(content);
+      blob = bb.getBlob(mimeType);
+    } else {
+      blob = new Blob([content], { type: mimeType });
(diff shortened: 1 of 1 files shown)


skddc:
Wow, that hound can be annoying. Looks like there's exactly one line with double quotes at the beginning of the file. ;)

galfert:
Ok, I hope I made dem hounds happy now :)

skddc:
+1

src/wireclient.js
```diff
@@ -93,10 +93,19 @@
   }
 
   function readBinaryData(content, mimeType, callback) {
-    var blob = new Blob([content], { type: mimeType });
+    var blob;
+    global.BlobBuilder = global.BlobBuilder || global.WebKitBlobBuilder;
```
silverbucket: Does the resulting var need to be global? This *might* cause issues with other libs, no?

skddc: Good point. Albeit if it re-assigns it to the prefixed one, there's no harm really, because it's the same thing.

skddc:
@silverbucket Is this +1 otherwise?

silverbucket:
Yup, +1

skddc:
Cool, thanks!

PR #868, opened 2015-06-19
When offline, ignore maxAge (with log) instead of failing

skddc:
This is a continuation of #861.

I reverted my changes to rejecting the promises for internal gets with maxAge and changed it to only set it to false when a default maxAge is added, when disconnected and offline. This is now just a small enhancement, as it was already done when disconnected, and now it's also ignored (with a message) when not explicitly defining a maxAge.

When you explicitly define one, the promise will be rejected, same as before, so you as a developer have the power over how you want it to behave. Either you care about it and set maxAge explicitly, or it just works when offline.

Ready for review/merge.


galfert:
+1

skddc:
@michielbdejong Maybe you could have a quick look, too?

michielbdejong:
I don't think there is one combination of defaults that everybody will find logical, so if this is the default behavior that all three of you (@skddc, @galfert , @silverbucket) like, then go for it.

The only thing I see missing in this PR is updating the documentation, I think whatever defaults you choose, the most important part is describing all situations (offline, connection timeout, other http request failures, and initial sync during page load vs. ajax request from a page that's already open) in a clear page on the wiki.

skddc:
> I don't think there is one combination of defaults that everybody will find logical, so if this is the default behavior that all three of you (@skddc, @galfert , @silverbucket) like, then go for it.

I'm proposing this change to everybody here, including @galfert and @silverbucket as well as anyone else using this lib or having contributed to it. I think @silverbucket is currently mostly offline while visiting family, so I don't expect timely feedback from him, though.

> The only thing I see missing in this PR is updating the documentation

Good point. The API docs are actually almost correct about the current behaviour, except not being explicit about "network problems" including the lib being in normal offline mode. I would interpret network problems to mean unexpected problems when it's thinking your online, but the request fails somehow. I'll try to update the docs accordingly and maybe put sth on the wiki.

src/remotestorage.js
```diff
@@ -39,9 +39,11 @@
       var self = this;
       if (this.local) {
         if (maxAge === undefined) {
-          if (this.connected) {
+          if ((typeof this.remote === 'object') &&
```
untitaker: Not sure why this has to be that strict, boot test would suffice?

src/remotestorage.js
```diff
@@ -39,9 +39,11 @@
...
+               this.remote.connected && this.remote.online) {
             maxAge = 2*this.getSyncInterval();
           } else {
+            RemoteStorage.log('Not setting default maxAge, because remote is not connected or offline');
```
untitaker: I think it reads better like this:

>Not setting default maxAge, because remote is *offline or not connected*

untitaker:
:+1: on this PR.

skddc:
Thanks for your review, @untitaker!

PR #871, opened 2015-06-28
Add optional token argument to connect() function

skddc:
Sometimes, e.g. when using the library from node.js, it's useful to be able to use a known token when connecting a storage and skip the authorization part altogether.

With this change, it's possible connect by calling the function directly and with a token (from a program or console), like so:

```js
remoteStorage.connect('basti@5apps.com', '123456789abcdef');
```

I'm not sure how to test this best, as there are currently no tests or mocks for WebFinger and Discovery, other than throwing a couple errors when e.g. the user address doesn't look like one. That means implied auth is also untested at the moment, which is not ideal. However the code there is rather easy and straight-forward, and it's pretty hard to break it when only touching it once a year. So maybe it's fine for now.

### DIFF
--- a/src/remotestorage.js
+++ b/src/remotestorage.js
@@ -311,13 +311,13 @@
      * Parameters:
      *   userAddress - The user address (user@host) to connect to.
      *
-     * Discovers the webfinger profile of the given user address and initiates
+     * Discovers the WebFinger profile of the given user address and initiates
      * the OAuth dance.
      *
      * This method must be called *after* all required access has been claimed.
      *
      */
-    connect: function (userAddress) {
+    connect: function (userAddress, token) {
       this.setBackend('remotestorage');
       if (userAddress.indexOf('@') < 0) {
         this._emit('error', new RemoteStorage.DiscoveryError("User address doesn't contain an @."));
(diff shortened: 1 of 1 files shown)


src/remotestorage.js
```diff
@@ -341,13 +341,20 @@
         this.remote.configure(info);
         if (! this.remote.connected) {
           if (info.authURL) {
-            this.authorize(info.authURL);
+            if (typeof token === 'undefined') {
```
Ragnis: I think it would make more sense to swap the if and else parts - check if the token is valid (a string?)
and otherwise continue with the authorization sequence.

skddc: Hmm. The overwhelming amount of cases is the first one, though. That will be the case for more than 99.99% of calls. So that's why I thought the skipping should be the `else`. Maybe with this explanation it seems more reasonable to you as well?

Ragnis: I still think that the type of `token` should be validated.

skddc: Ah, right. How about `else if typeof token === string else throw invalid`?

Ragnis: Yes

skddc: Cool. Will prepare it that way then. Thanks for the feedback! Much appreciated.

skddc:
@Ragnis Pushed a new version...

Ragnis:
Okay, looks good, +1.

silverbucket:
+1

skddc:
Yay, welcome back, @silverbucket! Hope you had a nice vacation. :)

skddc:
Merged and squashed locally.

PR #872, opened 2015-06-29
Fix syntax errors in strict mode

skddc:
Needed e.g. when running rs.js via babel-node.


Ragnis:
+1

skddc:
Merged/squashed locally. Didn't wait for second +1 as everybody is on vacation and this is a small fix.

PR #880, opened 2015-08-06
Added an oauth flow for cordova using InAppBrowser

lewisl9029:
Here's a quick change I made to get remoteStorage to play well with Cordova. 

It's working for my app but I may be missing some corner cases. 

No new tests are failing as far as I can tell (there were some console errors when I ran the tests, both before and after my changes, but the summary says 0 failures). Let me know if you want to see any new tests written for this.

I referred heavily to the Dropbox Oauth implementation in ng-cordova-oauth:

https://github.com/nraboy/ng-cordova-oauth/blob/master/src/oauth.js

I made a small modification to reuse the existing extractParams function, and changed the redirectUri and clientId to be compatible with Cordova. 

I'm not sure what we can use for clientId in Cordova though. Nothing seems to be distinctive enough in window.location under Cordova (I've included a snapshot of what window.location looks like in my cordova app). Would it be a good idea to ask app developers for a clientId as an input for their app instead of generating it?

### DIFF
--- a/src/authorize.js
+++ b/src/authorize.js
@@ -1,12 +1,12 @@
 (function (global) {
 
-  function extractParams() {
+  function extractParams(url) {
     //FF already decodes the URL fragment in document.location.hash, so use this instead:
-    var location = RemoteStorage.Authorize.getLocation(),
-        hashPos  = location.href.indexOf('#'),
+    var location = url || RemoteStorage.Authorize.getLocation().href,
+        hashPos  = location.indexOf('#'),
         hash;
     if (hashPos === -1) { return; }
-    hash = location.href.substring(hashPos+1);
+    hash = location.substring(hashPos+1);
     // if hash is not of the form #key=val&key=val, it's probably not for us
     if (hash.indexOf('=') === -1) { return; }
     return hash.split('&').reduce(function (m, kvs) {
@@ -37,6 +37,27 @@
       url += '&state=' + encodeURIComponent(redirectUri.substring(hashPos+1));
     }
     url += '&response_type=token';
+
+    if (global.cordova) {
+      return RemoteStorage.Authorize.openWindow(
+          url,
+          redirectUri,
+          'location=no,clearsessioncache=yes,clearcache=yes'
+        )
+        .then(function (authResult) {
+          remoteStorage.remote.configure({
+            token: authResult.access_token
+          });
+
+          // sync doesnt start until after reload
+          // possibly missing some initialization step?
+          global.location.reload();
+        })
+        .then(null, function (error) {
+          console.error(error);
+        });
+    }
+
     RemoteStorage.Authorize.setLocation(url);
   };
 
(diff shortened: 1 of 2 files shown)


src/authorize.js
```diff
@@ -37,6 +37,27 @@
...
+
+    if (global.cordova) {
+      return RemoteStorage.Authorize.openWindow(
+          url,
+          redirectUri,
+          'location=no,clearsessioncache=yes,clearcache=yes'
+        )
+        .then(function (authResult) {
+          remoteStorage.remote.configure({
+            token: authResult['access_token']
```
lewisl9029: Ok I'm not sure what I should do about this now...

I had authResult.access_token before and it complained that it's not in camelCase. =/

skddc: Just ignore it. Sometimes the hound is an idiot. ;)

silverbucket:
This is great, thanks @lewisl9029 ... I don't have any experience with Cordova personally, but from reading over the code I think it looks pretty solid. There's one are where you have some comments, does that need to be resolved before this pull request is ready? What are the implications of that issue?

skddc:
> There's one are where you have some comments

@silverbucket typo?

silverbucket:
@skddc yeah, edited my post

skddc:
Exciting! Good job.

> I'm not sure what we can use for clientId in Cordova though. Nothing seems to be distinctive enough in window.location under Cordova (I've included a snapshot of what window.location looks like in my cordova app). Would it be a good idea to ask app developers for a clientId as an input for their app instead of generating it?

We could also try to draw some information from http://docs.phonegap.com/en/edge/cordova_device_device.md.html -- not sure if it's already enough. Maybe it would be good to use in addition to a custom ID, so that you can revoke access per device.

galfert:
Nice. Very happy to see you got it working.

> We could also try to draw some information from http://docs.phonegap.com/en/edge/cordova_device_device.md.html -- not sure if it's already enough. Maybe it would be good to use in addition to a custom ID, so that you can revoke access per device.

Maybe the app's bundle identifier is also a good candidate to include. Although I don't even know from the top of my head if you can access that one from the app.

lewisl9029:
There are Cordova plugins for accessing the bundle identifier, but there's no guarantee that any given app would have it included.

I ended up just replacing the long comments with a reminder to eventually replace that part with a custom clientId, since it feels like that part is probably a different feature entirely.

silverbucket:
+1

skddc:
So, I thought about this for a bit, and we cannot use the `client_id` with a custom string, because in order to make RS connect a bit more secure, we (at @5apps) automatically set the `client_id` to the `redirect_uri` host. So do other servers/providers.

We could use a custom new parameter, like e.g. "display_name", so that the user is still shown the actual client_id in addition to a human-readable app name. That'd be useful for normal connects as well.

lewisl9029:
I have redirect_uri hard-coded to a dummy url (`'http://localhost/callback'`) for Cordova because it shouldn't matter where we redirect to, since we extract parameters from the InAppBrowser window and close it before the page even loads.

If we base the client_id on this redirect_uri then the current implementation would result in having the same client_id for all Cordova apps. Not sure if that'd be acceptable.

skddc:
Yes, that's the problem with localhost. We need to solve that somehow. As proposed in the original issue, an app can use its own domain and direct the DNS to `127.0.0.1`. You could test it with @galfert's `lolcathost.de`.

Try using `yourappname.lolcathost.de` for example.

lewisl9029:
To do that though, we would first need to get the `yourappame` part from the app developer as an input somehow, correct?

skddc:
The whole redirect URI in that case. The developer needs to set it up with their domain. Something like `cordova.myapp.com` for example.

lewisl9029:
Alright sounds good.

Would you like me to try implementing that as a part of this PR or do you think it should be spun-off as a separate feature?

skddc:
I think we should solve the whole Cordova flow with this one. I'll gladly help.

lewisl9029:
I just pushed some changes to allow for a custom cordovaRedirectUri in remoteStorage.connect(). 

I just tested it with `"https://toc.im"` and in both the OAuth dialog and the 5apps.com storage panel I do see `toc.im` as the id.

Seems to be working as expected.

@skddc 
I don't think the localhost redirect would be necessary.

In the current Cordova flow, we launch a new InAppBrowser window for OAuth, then extract the tokens from the InAppBrowser URL as soon as we see an url matching the redirect URI, and then immediately close the InAppBrowser window to return to the app. The redirect URI shouldn't ever finish loading.

skddc:
Oh, right! Good idea. It's of course still useful to have a unique host for a Cordova app, so that authorizations are unique to the app and don't get mixed up. We can put that in the docs.

IIRC we once had normal support for a `redirectUri` param with `displayWidget`. I even found it being used in my Ember apps, but it doesn't seem to work anymore and code and docs appear to not contain it. Was probably lost during the 0.8 rewrite.

I think we should re-introduce it as a general param.

@lewisl9029 I'll give you push access for this org, so you can push your branch to this repo. That way we can start jumping in and editing the same code/branch as you, so we can finish it up together.

lewisl9029:
Thanks @skddc, I just pushed the branch. 

Will we have to close this and open a PR for the new branch or is there some way of modifying this PR?

skddc:
Yes, we'll have to continue in a new one.

skddc:
Continued in #881.

PR #881, opened 2015-08-08
Added an oauth flow for cordova using InAppBrowser

lewisl9029:
Continuing from #880.

### DIFF
--- a/src/authorize.js
+++ b/src/authorize.js
@@ -42,7 +42,7 @@
       return RemoteStorage.Authorize.openWindow(
           url,
           redirectUri,
-          'location=no,clearsessioncache=yes,clearcache=yes'
+          'location=yes,clearsessioncache=yes,clearcache=yes'
         )
         .then(function (authResult) {
           remoteStorage.remote.configure({
--- a/src/authorize.js
+++ b/src/authorize.js
@@ -44,17 +44,19 @@
           redirectUri,
           'location=yes,clearsessioncache=yes,clearcache=yes'
         )
-        .then(function (authResult) {
+        .then(function(authResult) {
           remoteStorage.remote.configure({
             token: authResult.access_token
           });
 
-          // sync doesnt start until after reload
+          // TODO
+          // sync doesn't start until after reload
           // possibly missing some initialization step?
           global.location.reload();
         })
-        .then(null, function (error) {
+        .then(null, function(error) {
           console.error(error);
+          remoteStorage.widget.view.setState('initial');
         });
     }
 
@@ -101,7 +103,7 @@
   };
 
   /**
-   * Open new InAppBrowser window for Oauth on cordova
+   * Open new InAppBrowser window for OAuth in Cordova
    */
   RemoteStorage.Authorize.openWindow = function (url, redirectUri, options) {
     var pending = Promise.defer();
--- a/src/authorize.js
+++ b/src/authorize.js
@@ -1,12 +1,12 @@
 (function (global) {
 
-  function extractParams() {
+  function extractParams(url) {
     //FF already decodes the URL fragment in document.location.hash, so use this instead:
-    var location = RemoteStorage.Authorize.getLocation(),
-        hashPos  = location.href.indexOf('#'),
+    var location = url || RemoteStorage.Authorize.getLocation().href,
+        hashPos  = location.indexOf('#'),
         hash;
     if (hashPos === -1) { return; }
-    hash = location.href.substring(hashPos+1);
+    hash = location.substring(hashPos+1);
     // if hash is not of the form #key=val&key=val, it's probably not for us
     if (hash.indexOf('=') === -1) { return; }
     return hash.split('&').reduce(function (m, kvs) {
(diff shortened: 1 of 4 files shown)


skddc:
Continuation of #880.

src/authorize.js
```diff
@@ -37,16 +37,40 @@
       url += '&state=' + encodeURIComponent(redirectUri.substring(hashPos+1));
     }
     url += '&response_type=token';
+
+    if (global.cordova) {
```
skddc: I think this can also be `phonegap`. Not sure if that has `cordova` then. @galfert Do you know that?

src/authorize.js
```diff
@@ -1,12 +1,12 @@
 (function (global) {
 
-  function extractParams() {
+  function extractParams(url) {
     //FF already decodes the URL fragment in document.location.hash, so use this instead:
-    var location = RemoteStorage.Authorize.getLocation(),
-        hashPos  = location.href.indexOf('#'),
+    var location = url || RemoteStorage.Authorize.getLocation().href,
+        hashPos  = location.indexOf('#'),
```
skddc: Did you see the comment about why it was using `href` above?

lewisl9029: Yes, I saw the comment.
I'm not sure if it has anything to do with my changes though? It seems to me that this is functionally identical to the previous implementation when no `url` is passed. 

I might have missed your point.

skddc: Oh damn, I didn't see the `.href` in the other line. :)

Never mind!

skddc: Oh damn, I didn't see the `.href` in the other line. :)

My bad, never mind!

src/authorize.js
lewisl9029: I've seen `window.cordova || window.PhoneGap || window.phonegap` used in some places. That's probably the safest option for legacy support.

Then again, https://github.com/nraboy/ng-cordova-oauth/blob/master/src/oauth.js uses just `window.cordova` and nobody has complained so far.

I'll leave the final decision to you guys.

lewisl9029:
Just realized we probably needed a check for the InAppBrowser plugin because none of this will work without it.

skddc:
So, I just got this to work in a new app that might eventually become a contacts sync app. I used the widget in that one for now, which is why I adjusted workflow to set the token via a property on the remoteStorage instance instead of an argument to the connect function.

I think we should probably introduce a setter function for this, similar to `setBackgroundSyncInterval` et al. Maybe even a generic `remoteStorage.setConfig(key, value)`. Anyone has opinions on that?

/cc @Ragnis @silverbucket @galfert

There are also no tests for the new functionality yet. I could try to add some soon.

skddc:
Here's a proof of concept Ember app that has a working connect:

https://github.com/skddc/cordova-rs-sync

Redirect URI config happens here atm: https://github.com/skddc/cordova-rs-sync/blob/master/app/initializers/remotestorage.js#L7

skddc:
Added a setter function for the redirect URI. In your app, you can/should now do this:

```js
remoteStorage.setCordovaRedirectUri('https://android.myawesomeapp.com');
```

I also pushed an updated example app.

skddc:
FYI: deactivated that nagging hound.

skddc:
There's one thing left to do: when you close the in-app browser window, either by using the X or hitting the back button on Android, the lib stays in "connecting" mode. I'll check if it's a general bug, or if we have to fix it for Cordova only.

skddc:
@lewisl9029 Wanna have a look at my changes? Also, let me know if you want to try your hand at the OAuth abortion thing, or if I should have a look at that.

src/authorize.js
```diff
@@ -37,16 +37,40 @@
...
+          remoteStorage.remote.configure({
+            token: authResult.access_token
+          });
+
+          // sync doesnt start until after reload
+          // possibly missing some initialization step?
+          global.location.reload();
+        })
+        .then(null, function (error) {
+          console.error(error);
```
lewisl9029: Need to change lib state here in addition to (or instead of?) logging to fix the OAuth abortion issue. Not sure how that should be done though.

lewisl9029:
@skddc I've been following your changes and they looked great. =)

I left a comment at where I think we need to add more error handling to fix the OAuth abortion issue, but I'm not sure how to actually get remoteStorage lib to change state. Would you mind taking a look and adding that?

src/authorize.js
skddc: Thanks! I'll check it out.

skddc:
So, that last commit solved it I think.

We need a second person reviewing this PR now. /cc @galfert @Ragnis @silverbucket

skddc:
Ping reviewers. Let's get this merged.

skddc:
If you have an Android phone at hand, you can test this branch with my test app using the following install package file:

https://storage.5apps.com/basti/public/shares/150902-1553-android-debug.apk

galfert:
I tried @skddc's app on iOS. Connecting works fine. But when closing the in-app browser before finishing the OAuth dance the lib still stays in "connecting" state.

galfert:
Ha, I just found out why it wasn't working for me. @skddc's example app didn't use the most recent build of this PR.

Now everything is working fine.

src/authorize.js
```diff
@@ -37,16 +37,42 @@
...
+          remoteStorage.remote.configure({
+            token: authResult.access_token
+          });
+
+          // TODO
+          // sync doesn't start until after reload
+          // possibly missing some initialization step?
+          global.location.reload();
+        })
+        .then(null, function(error) {
```
galfert: Why add a second `then` with just the error handler instead of adding the handler to the first `then`?

untitaker: Possibly for readability, although you can use `catch` for that too?

lewisl9029: @untitaker `.catch` is what I'd normally use, but I only remember seeing `.then(null,` used elsewhere in the project so I thought that might have been a deliberate style choice and followed it.

@galfert I believe adding a handler to the previous `.then` will cause it to miss any errors triggered in the success callback of the same `.then`. Adding a dedicated error handler at the end of the chain is safer in that sense.

galfert: Thanks, that makes sense.

galfert:
+1

skddc:
Great! Updated the changelog and merging now. Also opening an issue for the sync not starting.

Thanks a lot, @lewisl9029 for contributing this feature and everyone for reviewing and testing!

lewisl9029:
Thanks to you guys too for building this awesome project. 

Glad to have been able to help! =D

Issue #882, opened 2015-08-27
IndexedDB getting into inconsistent state


galfert:
Sometimes the IndexedDB database gets into an invalid state, resulting into several different exceptions thrown when trying to open the database.

E.g. `The operation failed because the requested database object could not be found.`

skddc:
It should be noted that there's a fix for this being tested in the wild right now. So no need for anyone else to pick up the issue.

PR #886, opened 2015-09-07
Reset IndexedDB database when opening failed (refs #882)

galfert:
This PR catches errors on opening the IndexedDB database and then resets the database in order to get it back into a consistent state.

We have included this fix in a heavily used app since over a week and haven't seen any of the errors mentioned in #882 since.

### DIFF
--- a/src/indexeddb.js
+++ b/src/indexeddb.js
@@ -268,33 +268,55 @@
       callback("timeout trying to open db");
     }, 10000);
 
-    var req = indexedDB.open(name, DB_VERSION);
+    try {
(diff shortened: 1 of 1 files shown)


skddc:
I already reviewed this for use in our app. Would be great if @Ragnis and/or @silverbucket could have a another look at it. It's fairly simple and unfortunately also not really possible to test, so a quick "code looks ok" and trusting in our production app would be sufficient in my opinion.

silverbucket:
Just read it over and it looks good to me. +1

skddc:
thanks-ht

Issue #888, opened 2015-09-11
Caching usability and handling of empty paths


untitaker:
I found the following line in your HelloWorld example:

    privateClient.cache("")

https://github.com/remotestorage/starter-kit/blob/master/apps/hello-world/notes.js

The `cache` method isn't documented, so I removed it and naively switched to:

    privateClient.storage.caching.enable("")

However, I didn't know that this cache interface wasn't relative to the client's root. In addition to making remoteStorage try to cache the universe, this line also made it send requests to the storage root without a trailing slash, which violates the spec.

skddc:
These lines are both weird. Are you referring to https://github.com/remotestorage/starter-kit maybe?

untitaker:
Exactly, I've linked a file from that repo in the OP. It's on the second line.

skddc:
Ah, missed it last night.

The starter kit has not really been peer-reviewed or maintained by the current rs.js core team. The normal method is `remoteStorage.caching.enable('/my-category/subdir/')` which makes it more obvious that it's relative to root. I've never even seen caching methods on the client I think. :)

It's documented here, but examples for `enable` are missing, and docs are sparse in general: https://remotestorage.io/doc/code/files/caching-js.html#RemoteStorage.Caching

I started a new guide on the wiki yesterday, which -- ideally with the help of you and others -- eventually I'd like to turn into the go-to resource for learning about the general usage of remoteStorage.js. In the end, it should describe every step of the way, including caching, conflict handling and all (maybe some of it on seperate pages, of course, but linked from there).

untitaker:
What about the undocumented client method I used? It seemed to work fine, and yet, if it isn't documented perhaps it should be removed?

On 12 September 2015 12:26:21 CEST, Sebastian Kippe <notifications@github.com> wrote:
>Ah, missed it last night.
>
>The starter kit has not really been peer-reviewed or maintained by the
>current rs.js core team. The normal method is
>`remoteStorage.caching.enable('/my-category/subdir/')` which makes it
>more obvious that it's relative to root. I've never even seen caching
>methods on the client I think. :)
>
>It's documented here, but examples for `enable` are missing, and docs
>are sparse in general:
>https://remotestorage.io/doc/code/files/caching-js.html#RemoteStorage.Caching
>
>I started a new guide on the wiki yesterday, which -- ideally with the
>help of you and others -- eventually I'd like to turn into the go-to
>resource for learning about the general usage of remoteStorage.js. In
[...]

PR #889, opened 2015-09-12
Reject more invalid paths for Caching.set()

untitaker:
See #888

### DIFF
--- a/src/caching.js
+++ b/src/caching.js
@@ -53,7 +53,16 @@
       if (typeof(path) !== 'string') {
         throw new Error('path should be a string');
       }
-      if (typeof(value) === 'undefined') {
+      if (!RemoteStorage.util.isFolder(path)) {
+        throw new Error('path should be a folder');
+      }
+      if (
+        this._remoteStorage &&
+        this._remoteStorage.access &&
+        !this._remoteStorage.access.checkPathPermission(path, 'r')) {
+        throw new Error('no access to this path. You first have to claim access to it.')
+      }
+      if (!(value === 'FLUSH' || value == 'SEEN' || value == 'ALL')) {
         throw new Error("value should be 'FLUSH', 'SEEN', or 'ALL'");
       }
 
@@ -141,6 +150,7 @@
      **/
     reset: function () {
       this._rootPaths = {};
+      this._remoteStorage = null;
     }
   };
 
@@ -160,6 +170,8 @@
     }
   });
 
-  RemoteStorage.Caching._rs_init = function () {};
+  RemoteStorage.Caching._rs_init = function (remoteStorage) {
+    this._remoteStorage = remoteStorage;
+  };
 
 })(typeof(window) !== 'undefined' ? window : global);
--- a/test/unit/caching-suite.js
+++ b/test/unit/caching-suite.js
@@ -57,6 +57,30 @@ define(['requirejs', 'fs'], function(requirejs, fs, undefined) {
       },
 
       {
+        desc: "#set() with empty string given throws an error",
+        run: function(env, test) {
+          try {
+            env.caching.set("");
+            test.result(false, "set() didn't fail");
+          } catch(e) {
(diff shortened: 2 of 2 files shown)


src/caching.js
```diff
@@ -160,6 +170,8 @@
     }
   });
 
-  RemoteStorage.Caching._rs_init = function () {};
+  RemoteStorage.Caching._rs_init = function (remoteStorage) {
+    this._remoteStorage = remoteStorage;
```
untitaker: Is this value already available anywhere?

test/unit/caching-suite.js
```diff
@@ -57,6 +57,18 @@ define(['requirejs', 'fs'], function(requirejs, fs, undefined) {
...
+        run: function(env, test) {
+          try {
+            env.caching.set("");
+            test.result(false, "set() didn't fail");
+          } catch(e) {
+            test.result(true);
+          }
+        }
+      },
+
```
skddc: Cool. Maybe add another one for a non-folder string.

src/caching.js
skddc: Unfortunately, loading/startup process is currently a bit inconsistent and (imo) hacky. In the same file, there's [a note](https://github.com/remotestorage/remotestorage.js/blob/master/src/caching.js#L147-L151) about how the instance is not available yet and that this should be changed eventually. I think @galfert and I added that last time we tried to understand it from just reading the code.

To be honest, I don't know much about it from the top off my head, because I have to re-read all of it every time in order to remember. :/

src/caching.js
```diff
@@ -53,7 +53,16 @@
       if (typeof(path) !== 'string') {
         throw new Error('path should be a string');
       }
-      if (typeof(value) === 'undefined') {
+      if (!RemoteStorage.util.isFolder(path)) {
+        //throw new Error('path should be a folder');
```
skddc: Any reason this is commented?

untitaker: I kind of forgot about this, but as said in #888 there are tests calling `Caching.set` with non-folder paths (trailing slash is missing). Should I just correct those tests?

untitaker:
Also added a commit that simplifies the Travis configuration, hope you don't mind it's in the same PR.

test/unit/caching-suite.js
untitaker: Done.

src/caching.js
skddc: Yer, that'd be great.

untitaker: Alright done, though I still wonder why there are errors flashing across the screen and yet the tests pass.

skddc:
> Also added a commit that simplifies the Travis configuration, hope you don't mind it's in the same PR.

Nice!

skddc:
Looks to me like this is ready to merge. Needs a second +1 though.

/cc @Ragnis @silverbucket @galfert @michielbdejong

untitaker:
Can I cherry-pick ed9aa0deb2be13b70de2ee90d57294e107b3a08d to master?

skddc:
Actually I just noticed that `npm test` doesn't contain the single suite runs, which we added intentionally, because some of them once failed while all at once would run without errors (due to some quirks with loading things in node.js/jaribu there). So I guess we should add those to `npm test` instead now.

untitaker:
@skddc Please check the latest commit, I think I've come up with a good solution.

untitaker:
Factored out the buildscript improvements into: https://github.com/remotestorage/remotestorage.js/pull/892

skddc:
lgtm

silverbucket:
+1 looks good to me, thanks @untitaker

skddc:
I'm adding some minor things and merging locally.

skddc:
Done. https://github.com/remotestorage/remotestorage.js/commit/26fb110683287d86be53df92b0a15134a2d8786b

Thanks, @untitaker!

PR #766, opened 2014-09-15
don't send change event if body nor contentType changed

michielbdejong:
fixes #764

* [x] fix for string data
* [x] write test for this with string data
* [x] fix for binary data
* [x] write test for this with binary data

PR #890, opened 2015-09-12
don't send change event if body nor contentType changed, fixes 764

untitaker:
Rebase of #766

### DIFF
--- a/src/cachinglayer.js
+++ b/src/cachinglayer.js
@@ -326,14 +326,19 @@
             delete nodes[path];
           }
           else if (isDocument(path)) {
-            changeEvents.push({
-              path:           path,
-              origin:         'window',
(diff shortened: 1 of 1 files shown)


untitaker:
Some commits from #766 have been cherry-picked into master, but then (probably accidentally) reverted by https://github.com/remotestorage/remotestorage.js/commit/88b8b45cc69d76de067d6bcb9c45fbe9bdd989ed. Even if the new behavior doesn't make it in, I think the commit restoring those tests should be cherry-picked.

untitaker:
I've moved those parts into a new PR: #891.

untitaker:
#766 is now the same as this.

skddc:
Alright, closed the other one. Thanks.

PR #891, opened 2015-09-12
Restore some tests

untitaker:
See #890


skddc:
+1

skddc:
As these are tests only I'll merge with only one +1.

Thanks, @untitaker!

Issue #903, opened 2015-10-15
Empty files leads to infinite fetching during sync process


untitaker:
This has the same root cause as https://github.com/remotestorage/remotestorage.js/issues/187

I wonder whether we should introduce a `node` type, with methods that do the rigorous checking for us. Currently the code of `sync.js` is littered with `typeof`-checks, unfortunately only partially.

skddc:
Good idea.

untitaker:
Where would I define such a type?

skddc:
It's a fairly complex topic, as you also have to implement it with the caching in mind.

untitaker:
It would only replace the `{...}` construction I see in `makeNode`. It's not even a real design change?

skddc:
Oh, ok. Haven't looked into it specifically. Just explore on your own.

raucao:
This one had long been fixed.

PR #904, opened 2015-10-19
WIP: Introduce Node type

untitaker:
This should include a fix for #903 when finished.

### DIFF
--- a/src/cachinglayer.js
+++ b/src/cachinglayer.js
@@ -12,15 +12,17 @@
    * talk to.
    */
 
+  var Node = RemoteStorage.util.Node;
   var isFolder = RemoteStorage.util.isFolder;
   var isDocument = RemoteStorage.util.isDocument;
   var deepClone = RemoteStorage.util.deepClone;
   var equal = RemoteStorage.util.equal;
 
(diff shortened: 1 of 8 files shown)


galfert:
What's left to do here?

Maybe you could add a todo style list, then other people might be able to help out as well.

untitaker:
IIRC i was only fixing testcase for testcase. However, I might have introduced so many bugs that it might be better to start from scratch.

The idea was to introduce a Node type instead of the ad-hoc object we currently use. The stronger typing would have prevented the bug linked in the OP, IMO

On 18 July 2016 18:45:26 CEST, Garret Alfert <notifications@github.com> wrote:
>What's left to do here?
>
>Maybe you could add a todo style list, then other people might be able
>to help out as well.
>
>---
>You are receiving this because you were assigned.
>Reply to this email directly or view it on GitHub:
>https://github.com/remotestorage/remotestorage.js/pull/904#issuecomment-233386414

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

untitaker:
I'm closing this because I currently don't have time to work on it.

Issue #908, opened 2015-10-29
do not set empty state parameter when fragment empty


fkooman:
When an application uses empty hashes, i.e. `https://example.org/index.html#` the state parameter in the authorize URI is set without a value, this is not allowed. See https://tools.ietf.org/html/rfc6749#appendix-A.5.

PR #909, opened 2015-10-29
Do not set empty state parameter in authorize URI

fkooman:
According to https://tools.ietf.org/html/rfc6749#appendix-A.5 an
empty state query parameter is not allowed. This can occur when
the authorize step is started when the application has a fragment
set, but no value in it, e.g. https://example.org/foo.html#.

Fix for issue #908

### DIFF
--- a/src/authorize.js
+++ b/src/authorize.js
@@ -33,7 +33,7 @@
     url += 'redirect_uri=' + encodeURIComponent(redirectUri.replace(/#.*$/, ''));
     url += '&scope=' + encodeURIComponent(scope);
     url += '&client_id=' + encodeURIComponent(clientId);
-    if (hashPos !== -1) {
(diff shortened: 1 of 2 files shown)


untitaker:
Behavior LGTM, but the whole parsing looks pretty "organically grown". Even [this](https://gist.github.com/jlong/2428561) looks nicer IMO :P

skddc:
> Behavior LGTM, but the whole parsing looks pretty "organically grown". Even this looks nicer IMO :P

Feel free to refactor. :)

skddc:
Rebased and merged locally. Thanks, @fkooman!

skddc:
@fkooman By the way, I added you to the collab team, but you're also already an org owner. So you can create and push directly to this repo in the future. ;)

Issue #919, opened 2015-12-06
Using RemoteStorage with webpack in a cordova app raise a undefined remoteStorage error


khertan:
Using RemoteStorage with webpack in a cordova app raise a undefined remoteStorage error at autorize.js:58

skddc:
Fix merged to master.

PR #920, opened 2015-12-06
Fix undefined remoteStorage when using it in cordova (#919)

khertan:
When using remoteStorage.js with webpack , the remoteStorage vars isnt defined as it s not anymore available in scope. and remoteStorage is use for cordova compatibility.

### DIFF
--- a/src/authorize.js
+++ b/src/authorize.js
@@ -25,7 +25,7 @@
     document.location = redirectUri;
   };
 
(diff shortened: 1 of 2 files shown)


skddc:
The Authorize function is also used in the Dropbox and Google Drive backends. I added the instance argument to those as well.

Thanks, @khertan!

PR #925, opened 2016-04-05
Publish docs to gh-pages

skddc:
Adds the same script that is used in the modules repo in order to publish to gh-pages. I already published the current state, so I can link it on the new site/wiki:

https://remotestorage.github.io/remotestorage.js

### DIFF
--- a/scripts/publish-docs.sh
+++ b/scripts/publish-docs.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+make doc && \
+rm -rf tmp && \
+cp -R doc/code tmp && \
+rm -rf doc/code && \
+git checkout doc/config && \
+git checkout gh-pages && \
+cp -R tmp/* . && \
+rm -rf tmp && \
+git add . && \
+git commit -a -m "Update docs - `date -u`" && \
+git push origin gh-pages && \
+git checkout master
--- a/scripts/publish-docs.sh
+++ b/scripts/publish-docs.sh
@@ -1,14 +1,15 @@
 #!/bin/bash
+set -ex
 
-make doc && \
-rm -rf tmp && \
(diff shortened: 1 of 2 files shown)


scripts/publish-docs.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
untitaker: Use `set -e` instead of the `&&` chain. You can also use `set -x` (combined `set -ex`) to print the commands before executing them.

skddc: So I would just add `-ex` to the end of the shebang line?

untitaker: Actually having `-ex` within the shebang line will also print all the commands from your `.bashrc` it seems. This works and doesn't execute `echo b`:

#!/bin/sh

set -ex

echo a
false
echo b

skddc: Seems to work well. Thanks a lot!

untitaker: You also probably can use any POSIX shell in the shebang other than bash.

Either way :+1:

On 7 April 2016 01:24:33 CEST, Sebastian Kippe <notifications@github.com> wrote:
>> @@ -0,0 +1,14 @@
>> +#!/bin/bash
>
>Seems to work well. Thanks a lot!
>
>---
>You are receiving this because you commented.
>Reply to this email directly or view it on GitHub:
>https://github.com/remotestorage/remotestorage.js/pull/925/files/627754e2f826033ad0e6737c38d95b7264875a9b#r58798898

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

galfert:
+1

skddc:
Thanks.

PR #926, opened 2016-04-05
Change dev server recommendation in README

skddc:
... and link to servers page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,12 +22,13 @@ servers and syncing data across devices and applications.
 
 To develop remoteStorage-enabled apps, you need to have a
 remoteStorage-compatible storage account. We recommend
-[reStore](https://github.com/jcoglan/restore) for running a local test server.
-(Use the latest version from GitHub, not npm!)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,12 +22,13 @@ servers and syncing data across devices and applications.
 
 To develop remoteStorage-enabled apps, you need to have a
 remoteStorage-compatible storage account. We recommend
-[reStore](https://github.com/jcoglan/restore) for running a local test server.
```
untitaker: Can't we just remove all recommendations from the README, and move them into the wiki?

skddc: What page though? The "servers" page isn't just focused on developers, so I think the README that's contained in the repo is a good place to give that pointer. The next paragraph there points to the wiki.

untitaker: Good point about dev-specific recommendation.

untitaker:
:+1:

README.md
skddc: Also good point about the wiki, though. :) I think we should replicate that at least in the beginner's guide, if it isn't there yet.

galfert:
+1

PR #928, opened 2016-04-07
Put all files into a common prefix in DropBox

Ragnis:


skddc:
Welcome back, @Ragnis! :)

Looks like there are two failing Dropbox tests left to fix. Do you need help with it or wanna solve it on your own?

skddc:
Any news on this one?

Ragnis:
I haven't been able to get the tests running locally in order to fix them.

src/dropbox.js
```diff
@@ -41,7 +41,28 @@
   var hasLocalStorage;
   var AUTH_URL = 'https://www.dropbox.com/1/oauth2/authorize';
   var SETTINGS_KEY = 'remotestorage:dropbox';
-  var cleanPath = RS.WireClient.cleanPath;
+  var PATH_PREFIX = '/remotestorage';
+
+  /**
+   * Function: mapPath(path)
```
galfert: The documentation mentions a different function name than what's actually used (`mapPath` vs. `getDropboxPath`).

skddc: Why not just fix it real quick?

galfert:
Testing in [Webmarks](https://github.com/skddc/webmarks), this doesn't work for me. When I create a new Bookmark, I see a `PUT` request that responds with a 200. I can see the folders (`remotestorage/bookmarks/archive`) have been created in my Dropbox folder, but there are no files in there. There seems to be a `DELETE` request shortly after the `PUT` that deletes the file again.

galfert:
I think I found the culprit. After the `PUT` it detects a conflict, because the path of the response doesn't correspond with the actual path here: https://github.com/remotestorage/remotestorage.js/blob/feature/796-dropbox-common-base-path/src/dropbox.js#L796.

galfert:
I fixed the comparison, but didn't have time yet to also write a test for it.

`PUT` requests work for me now.

But when clicking on the Dropbox icon in the widget, I see this error in the console:

> remotestorage.js:13440 Uncaught TypeError: rs.sync.bind is not a function

The error happens here: https://github.com/remotestorage/remotestorage.js/blob/feature/796-dropbox-common-base-path/src/dropbox.js#L854

galfert:
> I fixed the comparison, but didn't have time yet to also write a test for it.

Just saw that there are actually failing tests now. So it seems that there were already tests for this, but they were just not adjusted for the new paths yet.

galfert:
Fixed those tests.

galfert:
FYI: I found a general problem with the Dropxbox (and also GoogleDrive) initialization while working on #940.
When loading the app for the first time, the app sets the API keys only after the remoteStorage initialization has already happened. But the Dropbox backend only gets initialized, when there is an API key set.

So the widget shows the Dropbox icon, but the class hasn't been initialized, which results in an exception when clicking the dropbox icon.

When loading the app for the second time and the API key is loaded from localStorage before the Dropbox initialization, the Dropbox backend object will be created properly.

galfert:
I created a new issue for the backend initialization bug at #941, it's not really related to this PR.

skddc:
Are you saying it's ready for review/merge?

galfert:
Ok, I fixed the remaining issue where Dropbox wouldn't hook itself up properly into the sync cycle.

I think this PR is ready for review/merge now.

skddc:
Here's a deployed version of myfavoritedrinks for easy testing:

https://mfd-dropbox.5apps.com/

I'll leave that online. Ping me if you want deploy access on 5apps for that app.

skddc:
LGTM and tested with the above app. I ran into some unrelated issues, but will test some more and create seperate GH issues if needed.

skddc:
@Ragnis wanna have another look at what @galfert added there and try it out yourself?

galfert:
Weird. Just noticed that Travis had 2 failing tests (which worked for me locally). I restarted the Travis build and now they are green `¯\_(ツ)_/¯`

src/dropbox.js
```diff
@@ -811,7 +837,7 @@
           return Promise.reject(new Error("Cannot delete '" + path + "': too many files involved"));
         }
 
-        if (resp.status === 200) {
+        if (resp.status === 200 || resp.status === 404) {
```
gregkare: Why is this now checking for 404 as well as 200? I think there should be a comment about this because it is not clear to me

gregkare:
LGTM, I also tested this and only had unrelated issues (the same that @skddc encountered) and created an issue for it: #943

I added a question

I think it would be a good idea not to hard-code the API URI everywhere (https://api.dropbox.com/1/) Should I create a separate issue for it?

galfert:
> I think it would be a good idea not to hard-code the API URI everywhere (https://api.dropbox.com/1/) Should I create a separate issue for it?

I created an issue for it at #944.

gregkare:
@galfert I think you missed my question (https://github.com/remotestorage/remotestorage.js/pull/928#discussion_r71178250). Does anyone else know about this change?

src/dropbox.js
skddc: It's a delete request, so if it tries to delete something that doesn't exist, that would result in a 404 I think.

gregkare: That makes sense, thanks

gregkare:
+1

Issue #885, opened 2015-09-06
Sync doesn't start automatically after calling `configure` with token


skddc:
There's an issue left with the Cordova integration, which is that sync doesn't start automatically. Also, @galfert and I saw a similar issue with an app outside of Cordova as well.

https://github.com/remotestorage/remotestorage.js/blob/master/src/authorize.js#L52-L55

skddc:
I think we can close this, since we merged #930, right @galfert?

galfert:
Yes, this should be fixed now.

skddc:
Hooray! (and /cc @lewisl9029)

![thumbs-up-daft-punk](https://cloud.githubusercontent.com/assets/842/15427124/262a5d4a-1e93-11e6-90e6-e8af91ee003c.gif)

PR #930, opened 2016-04-30
Fix disconnect / initialization

galfert:
refs #885

This bugfix is still work in progress.

It fixes problems when running `remoteStorage.disconnect()`, which prevented proper initialization and reconnection afterwards without reloading the Browser.

For some of the fixes there are still tests missing.

Also, there is still some timing issue which sometimes prevents the proper starting of the sync after calling configure (as described in the referenced issue).

### DIFF
--- a/src/authorize.js
+++ b/src/authorize.js
@@ -48,11 +48,6 @@
           remoteStorage.remote.configure({
             token: authResult.access_token
           });
-
-          // TODO
-          // sync doesn't start until after reload
-          // possibly missing some initialization step?
-          global.location.reload();
         })
         .then(null, function(error) {
           console.error(error);
--- a/src/eventhandling.js
+++ b/src/eventhandling.js
@@ -39,7 +39,7 @@
     _emit: function (eventName) {
       this._validateEvent(eventName);
       var args = Array.prototype.slice.call(arguments, 1);
-      this._handlers[eventName].forEach(function (handler) {
+      this._handlers[eventName].slice().forEach(function (handler) {
         handler.apply(this, args);
       });
     },
--- a/src/i18n.js
+++ b/src/i18n.js
@@ -40,7 +40,9 @@
 
     setDictionary: function (newDictionary) {
       dictionary = newDictionary;
-    }
+    },
 
+    _rs_init: function() {
+    }
   };
 })();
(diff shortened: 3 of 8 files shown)


skddc:
I reviewed the code changes and tested a build with Webmarks. LGTM! To be honest, I don't know if there's much more we can unit-test there for now.

Do you know about cause of the timing issue already?

galfert:
I added some more tests and fixed the actual problem of remoteStorage not syncing after configuring/connecting the remote.

The problem was that sync only started automatically when the remote is already configured at the time the `Sync` instance gets created. This is always the case when either (re-)loading an app when already connected (because the data gets directly loaded from localStorage), or when using the OAuth dance via the widget (there the remote gets configured before `Sync` gets initialized after the redirect).

Now, `Sync` listens on the `connected` event and starts syncing then.

This PR is ready for review/test now.

src/authorize.js
```diff
@@ -48,11 +48,6 @@
           remoteStorage.remote.configure({
             token: authResult.access_token
           });
-
-          // TODO
-          // sync doesn't start until after reload
-          // possibly missing some initialization step?
-          global.location.reload();
```
skddc: great-success

skddc:
The code looked good to me at first, but when I tested it with Webmarks, I found that the connected and disconnected events no longer work correctly. Here's a video demontrating it:

https://storage.5apps.com/basti/public/shares/160506-2059-rsjs-test-1.webm

(Don't try to skip back and forth, we need to implement content-range requests in 5apps Storage ;))

galfert:
Thanks for testing. I'll try to find out what's wrong there.

galfert:
FYI: I couldn't reproduce this in Chrome, but I'm seeing the same behavior in Firefox. Currently trying to find out what's wrong here.

galfert:
Took a while to find the cause, but I think I got that sucker :)

Reason was that the list of event handlers was modified while the event emitter was iterating over it, and thus skipped over one entry.

I think this should all finally be good to go. Thanks again for pointing out the issue.

skddc:
Nice catch! Will test again asap.

skddc:
Works now and lgtm. Good job!

silverbucket:
Looks good, I reviewed the code and did some basic tests and didn't run into any problems. +1

galfert:
Thanks for testing and reviewing, guys.

PR #933, opened 2016-05-06
Fix getListing not rejecting promise when using illegal path

### LINKED ISSUE #932: getListing misbehaving when calling path without trailing slash
Instead of rejecting the promise, it resolves it with an empty object, throws an error and still does the get request.

Example:

```js
remoteStorage.scope('/bookmarks/').getListing('foobar').then(
  function(result) { console.log('promise resolved:', result) },
  function(error) { console.log('promise rejected') }
);
```

This logs `promise resolved: Object {}` and makes a `GET` request to `/bookmarks/foobar`, but also throws an error

> Possibly unhandled Error: Not a folder: foobar

galfert:
fixes #932 

Fixed by actually returning the rejected promise.

There even was a test that checked that `getListing` fails, but it failed for a different reason (network request didn't work).

I also cleaned up some debugging console logs from the test output.

### DIFF
--- a/src/baseclient.js
+++ b/src/baseclient.js
@@ -219,7 +219,7 @@
       if (typeof(path) !== 'string') {
         path = '';
       } else if (path.length > 0 && path[path.length - 1] !== '/') {
-        Promise.reject("Not a folder: " + path);
+        return Promise.reject("Not a folder: " + path);
       }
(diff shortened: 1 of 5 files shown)


skddc:
+1 lgtm

untitaker:
👍

Issue #506, opened 2013-11-28
Use Google Drive's appdata to sync application-owned data


rakyll:
Application data should not be visible to the user, Google Drive provides a hidden folder for each app to be able to sync application owned data such as preferences. Rather than syncing with a regular folder, use `appdata`.

More details are on https://developers.google.com/drive/appdata

PR #935, opened 2016-05-17
Bugfix/934-unresolved promise

### LINKED ISSUE #934: Problem when trying to get an unexisting remote file [cache + IndexedDB]
I have a problem when trying to get an unexisting remote file where the associated promise never gets resolved. This behavior only appears  with the "cached" version, the "no-cache" works fine.
I tracked the problem down to [cachinglayer.js#L109](https://github.com/remotestorage/remotestorage.js/blob/462ecc1e01e8933244fa1b4366ee440e159b4c72/src/cachinglayer.js#L109). Looks to me like the ```isOutdated``` func always returns ```true``` for unexisting remote files (I guess there is no information in the indexeddb), so apparently this means the condition will never return 404 at that point. May be later on there is some problem with the resolution of ```queueGetRequest``` and the synchronization process for a "not found" file... I didn't look further into it, was hopping this gives you enough hints to solve it, but I can investigate further if needed. Thanks a lot.

esroyo:
fixes #934 

The problem appears when trying to get a remote unexisting file (on a subdirectory, with cache), then the associated promise is not resolved.

The reason is that the callback associated to the task is removed in the function ```deleteChildPathsFromTasks``` because a task to get the parent exists.

[Edit] Forgot to comment the solution though it is obvious: before deleting the child paths from the tasks, we pick the associated callbacks and add them to the parent path.

Seems to work fine now.

Also fixed what looks to me like a little slip in the ```_timeStarted``` assignment.

### DIFF
--- a/src/sync.js
+++ b/src/sync.js
@@ -286,7 +286,7 @@
 
         for (var i=1; i<paths.length; i++) {
           if (this._tasks[paths[i]]) {
-            if (this._tasks[path]) {
+            if (this._tasks[path].length) {
               Array.prototype.push.apply(
                 this._tasks[paths[i]],
                 this._tasks[path]
--- a/src/sync.js
+++ b/src/sync.js
@@ -286,7 +286,8 @@
 
         for (var i=1; i<paths.length; i++) {
           if (this._tasks[paths[i]]) {
-            if (this._tasks[path].length) {
+            // move pending promises to parent task
+            if (Array.isArray(this._tasks[path]) && this._tasks[path].length) {
               Array.prototype.push.apply(
                 this._tasks[paths[i]],
                 this._tasks[path]
--- a/src/sync.js
+++ b/src/sync.js
@@ -961,7 +961,7 @@
       }
       for (path in this._tasks) {
         if (!this._running[path]) {
-          this._timeStarted = this.now();
+          this._timeStarted[path] = this.now();
           this._running[path] = this.doTask(path);
           this._running[path].then(this.finishTask.bind(this));
           numAdded++;
--- a/src/sync.js
+++ b/src/sync.js
@@ -286,6 +286,12 @@
 
         for (var i=1; i<paths.length; i++) {
           if (this._tasks[paths[i]]) {
+            if (this._tasks[path]) {
+              Array.prototype.push.apply(
+                this._tasks[paths[i]],
+                this._tasks[path]
+              );
+            }
             delete this._tasks[path];
           }
         }
--- a/test/unit/sync-suite.js
+++ b/test/unit/sync-suite.js
@@ -249,16 +249,20 @@ define(['bluebird', 'test/helpers/mocks', 'requirejs'], function(Promise, mocks,
           });
         }
       },
-
       {
         desc: "collectRefreshTasks gives preference to caching parent",
         run: function(env, test) {
           var tmpForAllNodes = env.rs.local.forAllNodes;
           var tmpNow = env.rs.sync.now;
+          var fakeCallback = function() {};
 
           test.assertAnd(env.rs.sync._tasks, {});
           test.assertAnd(env.rs.sync._running, {});
 
+          env.rs.sync.addTask(
+            '/foo/ba/and/then/some/sub/path',
+            fakeCallback // should be passed to the ancestor when overruled
+          );
           env.rs.sync.now = function() {
             return 1234568654321;
           };
@@ -301,7 +305,7 @@ define(['bluebird', 'test/helpers/mocks', 'requirejs'], function(Promise, mocks,
 
           env.rs.sync.collectRefreshTasks().then(function() {
             test.assertAnd(env.rs.sync._tasks, {
-              '/foo/': [],
+              '/foo/': [fakeCallback], // inherited from task '/foo/ba/and/then/some/sub/path'
               '/read/access/': []
             });
             env.rs.local.forAllNodes = tmpForAllNodes;


galfert:
When reviewing your fixes in this PR I tried to reproduce the problem using the current master as well as the latest release. But I couldn't get the promise to not resolve.

I tested this using

```javascript
remoteStorage.scope('/some/path/').getObject('some-non-existing-file').then(
  function(object) { console.log('Resolved:', object); },
  function(error) { console.error('Rejected:', error); }
);
```

It resolves as soon as the network request returns with a 404, resulting in the output

> Resolved: undefined

Same when using `getFile` instead of `getObject`, which resolves with an object with all its properties being `undefined`:

```javascript
{data: undefined, contentType: undefined, revision: undefined}
```

The only way I could get the promise to not resolve nor reject was when taking the computer offline and doing the `getObject` call while the remoteStorage client hasn't gotten into offline mode yet. In that case, the promise only resolves after the computer gets online again. But that also still happens when using this PR.

@esroyo: Could you please post a little snippet of code that didn't work for you? Thanks!

esroyo:
I can't give much more light on this issue, I tested with something as simple as yours:
```javascript
remoteStorage.scope('/appdata/quick-moodboards/').getObject('muuu.json')   
.then(                                                                     
      function(object) { console.log('Resolved:', object); },              
      function(error) { console.error('Rejected:', error); }             
);
````
Chrome console output:
```
(x) Failed to load resource: the server responded with a status of 404 (OK)	https://storage.5apps.com/esroyo/appdata/quick-moodboards/muuu.json
```
And nothing else happens, never ever.

Oddly enough when wrapping the code in a setTimeout with 500ms, it resolves:
```
(x) GET https://storage.5apps.com/esroyo/appdata/quick-moodboards/muuu.json 404 (OK)	https://storage.5apps.com/esroyo/appdata/quick-moodboards/muuu.json:1
Resolved: undefined
```
I can't tell what exactly of my environment triggers this behavior. But It is consistent, always happens, and I'm sure the diagnostic about the callback of the task being removed is correct because I debugged all the flow step by step.

I'm sorry I can't be more helpful with this. Regards.

**Comment by galfert** [comment]
I put the code that I tested with online. Lets see if that one also does not resolve for you.

You can find it at https://rs-promise-test.5apps.com.

After connecting to your remoteStorage account, it should show a button "Load data". Please press that button and see if an alert dialog shows up.

**Comment by skddc** [comment]
One more idea: could it be a conflict with another `Promise` polyfill library perhaps?

**Comment by esroyo** [comment]
First: your test code already includes my fix! how can it be?
rs-promise-test

Second: What I try to achieve is immediate access to the file, it is an app config file that should be loaded on the app bootstrapping process. Thus, the code would look like:
```javascript
      remoteStorage.access.claim('appdata', 'rw');
      remoteStorage.displayWidget('remotestorage-connect', { redirectUri: window.location.href });
      remoteStorage.scope('/appdata/quick-moodboards/').getObject('muuu.json')
      .then(
        function(object) {
          console.log('Resolved:', object);
          window.alert('The promise was resolved.');
        },
        function(error) {
          console.error('Rejected:', error);
          window.alert('The promise was rejected.');
        }
      );
```
May be I misunderstood how RS works, but I thought I could work with it even without connecting.
Anyway, in case you are already connected from a previous visit, this should work, right?
Without the fix, this doesn't resolve nor throws any error, except the 404 notice.

This example confirms that there is no interference from other libraries.

galfert:
You are right, I included the wrong build of remotestorage.js that already included your code. I changed it to use the current master version now.

I also adapted the example to yours, immediately trying to access the object when the app loads. And still the promise gets resolved, both when connected and not connected.

esroyo:
Holy cow! now FF resolves, still Chrome doesn't:
rs-test2

esroyo:
Tried out the example at [browserling](https://www.browserling.com/browse/win/vista/chrome/43/https%3A%2F%2Frs-promise-test.5apps.com) with the same Chrome version and it always resolves. So this looks like a problem of my own Chrome :cry:

esroyo:
In the picture below you can see the situation that leads to the promise getting lost, because the task for `"/appdata/quick-moodboards/muuu.json"` is going to be deleted along with the associated callback.

However, in a working browser like FF, this never happens because ```deleteChildPathsFromTasks``` is never called in the first place. Why is that? Could be related to the IndexedDB state?

rs-promise3

skddc:
You should also wait for rs.js to tell you that it's ready via its events. In the example you're trying to retrieve the file right away without even waiting for it to emit the `ready` or `connected` event.

We'll have to improve the documentation for that I guess.

skddc:
@galfert Aren't you using Chrome as well?

galfert:
> @galfert Aren't you using Chrome as well?

Yes, tested in both Chrome and Firefox. Both resolve the promise for me.

skddc:
By the way, completely unrelated, but:

I think `appdata` is not a good category name for your app, as it is too generic. I think it should use sth like `moodboards` instead. That way it is both clear to users and developers what might be stored in it, and it's also less likely that someone uses it as an invitation to store completely unrelated app data of their own new app in the same one.

esroyo:
@skddc Problem persists even when wrapping the `getObject` call into an on `ready` event listener. I believe this must be related to the browser storage state. It doesn't make sense that the `deleteChildPathsFromTasks` is only called on my Chrome and not in any other fresh browser.

I realize this is a weird and pretty obscure situation, but could you please consider applying my patch anyway? Impartially looking the code It still doesn't make sense to delete tasks without doing something with the associated callbacks, don't you think?

About the "unrelated" path comment: I didn't find any common approach to store appdata config/state and decided to go with the idea commented on #506 using `/appdata/{{app_name}}/`, thus having a common root for all my apps. I'd store the data itself in a base directory called `/moodboards` in case of making a new module for the data structure. The main reason is that I wouldn't like to have each app polluting the root directory with a app-related-config directory. I really appreciate your suggestions, do you really believe this is wrong approach?

skddc:
> I much appreciate your suggestions, do you really believe this is wrong approach?

Yes, very much so. This way, you're forcing users to give every one of your apps permission to access every other app's data, without it being evident. And the auth dialog will just show them that they give access to "appdata", which is impossible to recognise as a data type/content, and more importantly remember later.

skddc:
> @skddc Problem persists even when wrapping the getObject call into an on ready event listener. I believe this must be related to the browser storage state. It doesn't make sense that the deleteChildPathsFromTasks is only called on my Chrome and not in any other fresh browser.

I definitely saw very weird and broken Web Storage situations in the wild myself before. I remember one case, where the browser was so broken, that I had to manually delete the respective files on disk in order to reset the storage back-ends.

skddc:
> I realize this is a weird and pretty obscure situation, but could you please consider applying my patch anyway?

Yes, of course. We just wanted to find out what's happening and why we haven't seen this bug before in order to understand what the best solution would be.

esroyo:
After clearing all browser data, I've managed to reproduce the bug again with the code below. Looks like it only happens when there is information about the given directory in the IndexedDB. So if we create a fake file in there before trying to get the unexisting file, the bug kind of appears:
```javascript
      remoteStorage.access.claim('appdata', 'rw');

      remoteStorage.displayWidget('remotestorage-connect', { redirectUri: window.location.href });

      remoteStorage.on('ready', function() {
        var client = remoteStorage.scope('/appdata/');
        client.declareType('fake', {"type": "object"});

        client.storeObject('fake', 'quick-moodboards/test.json', {test: 1});

        client.getObject('quick-moodboards/tst.json')
        .then(
          function(object) {
            console.log('Resolved:', object);
            window.alert('The promise was resolved.');
          },
          function(error) {
            console.error('Rejected:', error);
            window.alert('The promise was rejected.');
          }
        );
      });
```
@skddc thank you for the advice about the naming/scoping, I completely agree with the reasoning.

Again, thanks to both of you for the kind support. Regards.

galfert:
Ok, that's weird. I'm sure I also tried to reproduce it with a directory that already contained files, but I couldn't. 

Now I just tried your snippet and I finally saw the same situation as you. Not on every load though, but sometimes the promise doesn't resolve. So there definitely is some timing issue.

I will try to find out if your proposed change might have any other implications. But from first glance it looks good to me. Some input from @michielbdejong about this would be greatly appreciated.

src/sync.js
```diff
@@ -286,6 +286,12 @@
 
         for (var i=1; i<paths.length; i++) {
           if (this._tasks[paths[i]]) {
+            if (this._tasks[path]) {
```
michielbdejong: how can this if statement be false if `path` comes from line 284?

esroyo: You are right, I think this line should read:
```javascript
if (this._tasks[path].length) {
```

michielbdejong:
ok, the code makes sense to me - only remarks:

* it would be nice to have a unit test that fails in current master but passes with this fix, but if that's hard to add then maybe you can skip that
* add a comment at the top of your diff like `// move pending promises to parent task` or similar.

esroyo:
Thanks. PR updated. I've modified an already existing related test to include the check. It makes sense to me to test it there, but let me know if you prefer a separate test.

galfert:
Great. Now with the added test I think it's perfect. Couldn't find any other problems either.

+1

michielbdejong:
awesome! r+

esroyo:
Thanks a lot guys, my pleasure.

galfert:
I took the liberty to consolidate some of the commits before merging the PR with bbc8cdddc4584ab.

@esroyo Big thanks again for your effort and welcome to the remoteStorage contributors.

PR #940, opened 2016-06-22
Fix errors in Safari private browsing mode

### LINKED ISSUE #921: QuotaExceededError: DOM Exception 22 in iOS Safari private browsing mode
In Safari on iOS, when in private browsing mode, the library throws the following error, which prevents the app from starting at all:

```
QuotaExceededError: DOM Exception 22: An attempt was made to add something to storage that exceeded the quota.
```

The error is happening in the `setBackend` function.

galfert:
closes #921

In Safari private browsing mode, writing to locaStorage throws an `QuotaExceededError` error.

This change tries to set an item for checking if localStorage is available.

### DIFF
--- a/src/authorize.js
+++ b/src/authorize.js
@@ -23,7 +23,7 @@
 
         // remove rsDiscovery param
         stateValue = stateValue.slice(stateValue.indexOf('rsDiscovery='),
-                                      encodedData.lenth + 12);
+                                      encodedData.length + 12);
 
         if (stateValue.length > 0) {
           m['state'] = state;
--- a/src/authorize.js
+++ b/src/authorize.js
@@ -26,7 +26,7 @@
                                       encodedData.length + 12);
 
         if (stateValue.length > 0) {
-          params['state'] = state;
+          params['state'] = stateValue;
         }
       } else {
         params[decodeURIComponent(kv[0])] = decodeURIComponent(kv[1]);
--- a/src/authorize.js
+++ b/src/authorize.js
@@ -11,7 +11,27 @@
     if (hash.indexOf('=') === -1) { return; }
     return hash.split('&').reduce(function (m, kvs) {
       var kv = kvs.split('=');
-      m[decodeURIComponent(kv[0])] = decodeURIComponent(kv[1]);
+
+      if (kv[0] === 'state' && kv[1].match(/userAddress/)) {
+        // extract userAddress from the state param
+        var stateValue = decodeURIComponent(kv[1]);
+        var userAddress = stateValue.substr(stateValue.indexOf('userAddress='))
+                                    .split('&')[0]
+                                    .split('=')[1];
+
+        m['userAddress'] = userAddress;
+
+        // remove userAddress param
+        stateValue = stateValue.slice(stateValue.indexOf('userAddress='),
+                                      userAddress.lenth + 12);
+
+        if (stateValue.length > 0) {
+          m['state'] = state;
+        }
+      } else {
+        m[decodeURIComponent(kv[0])] = decodeURIComponent(kv[1]);
+      }
+
       return m;
     }, {});
   }
@@ -27,6 +47,12 @@
 
   RemoteStorage.Authorize = function (remoteStorage, authURL, scope, redirectUri, clientId) {
     RemoteStorage.log('[Authorize] authURL = ', authURL, 'scope = ', scope, 'redirectUri = ', redirectUri, 'clientId = ', clientId);
+
+    // keep track of the userAddress during redirect if we can't save it in localStorage
+    if (!remoteStorage.localStorageAvailable()) {
+      redirectUri += redirectUri.indexOf('#') > 0 ? '&' : '#';
+      redirectUri += 'userAddress=' + remoteStorage.remote.userAddress;
+    }
 
     var url = authURL, hashPos = redirectUri.indexOf('#');
     url += authURL.indexOf('?') > 0 ? '&' : '?';
@@ -153,6 +179,13 @@
         if (params.error) {
           throw "Authorization server errored: " + params.error;
         }
+
+        // userAddress came with the redirect, because it couldn't be
+        // saved in localStorage
+        if (params.userAddress && params.access_token) {
+          remoteStorage.connect(params.userAddress, params.access_token);
+        }
+
         if (params.access_token) {
           remoteStorage.remote.configure({
             token: params.access_token
--- a/src/authorize.js
+++ b/src/authorize.js
@@ -12,18 +12,18 @@
     return hash.split('&').reduce(function (m, kvs) {
       var kv = kvs.split('=');
 
-      if (kv[0] === 'state' && kv[1].match(/userAddress/)) {
-        // extract userAddress from the state param
+      if (kv[0] === 'state' && kv[1].match(/rsDiscovery/)) {
+        // extract rsDiscovery data from the state param
         var stateValue = decodeURIComponent(kv[1]);
-        var userAddress = stateValue.substr(stateValue.indexOf('userAddress='))
+        var encodedData = stateValue.substr(stateValue.indexOf('rsDiscovery='))
                                     .split('&')[0]
                                     .split('=')[1];
 
-        m['userAddress'] = userAddress;
+        m['rsDiscovery'] = JSON.parse(atob(encodedData));
 
-        // remove userAddress param
-        stateValue = stateValue.slice(stateValue.indexOf('userAddress='),
-                                      userAddress.lenth + 12);
+        // remove rsDiscovery param
+        stateValue = stateValue.slice(stateValue.indexOf('rsDiscovery='),
+                                      encodedData.lenth + 12);
 
         if (stateValue.length > 0) {
           m['state'] = state;
@@ -48,10 +48,18 @@
   RemoteStorage.Authorize = function (remoteStorage, authURL, scope, redirectUri, clientId) {
     RemoteStorage.log('[Authorize] authURL = ', authURL, 'scope = ', scope, 'redirectUri = ', redirectUri, 'clientId = ', clientId);
 
-    // keep track of the userAddress during redirect if we can't save it in localStorage
+    // keep track of the discovery data during redirect if we can't save it in localStorage
     if (!remoteStorage.localStorageAvailable()) {
       redirectUri += redirectUri.indexOf('#') > 0 ? '&' : '#';
-      redirectUri += 'userAddress=' + remoteStorage.remote.userAddress;
+
+      var discoveryData = {
+        userAddress: remoteStorage.remote.userAddress,
+        href: remoteStorage.remote.href,
+        storageApi: remoteStorage.remote.storageApi,
+        properties: remoteStorage.remote.properties
+      };
+
+      redirectUri += 'rsDiscovery=' + btoa(JSON.stringify(discoveryData));
     }
 
     var url = authURL, hashPos = redirectUri.indexOf('#');
@@ -180,10 +188,10 @@
           throw "Authorization server errored: " + params.error;
         }
 
-        // userAddress came with the redirect, because it couldn't be
+        // rsDiscovery came with the redirect, because it couldn't be
         // saved in localStorage
-        if (params.userAddress && params.access_token) {
-          remoteStorage.connect(params.userAddress, params.access_token);
+        if (params.rsDiscovery) {
+          remoteStorage.remote.configure(params.rsDiscovery);
         }
 
         if (params.access_token) {
--- a/src/authorize.js
+++ b/src/authorize.js
@@ -22,8 +22,7 @@
         params['rsDiscovery'] = JSON.parse(atob(encodedData));
 
         // remove rsDiscovery param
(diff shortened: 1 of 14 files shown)


skddc:
+1

silverbucket:
I tried with myfavoritedrinks:

1. checkout this branch:
    git checkout -b bugfix/921-fix_quota_exceeded_error origin/bugfix/921-fix_quota_exceeded_error

2. make all

3. go to myfavoritedrinks checkout and copy over the built remotestorage.js

4. run local http server

5. browse to local address and log in.

6. When I come back to the app, and try to add a drink I get this error:

```
[Error] TypeError: null is not an object (evaluating 'backendName')
	activateBackend (remotestorage.js:6892)
	initial (remotestorage.js:6681)
	display (remotestorage.js:6647)
	display (remotestorage.js:6281)
	displayWidget (remotestorage.js:6348)
	init (app.js:21)
```

I very well could be doing something wrong (it's been a while), which is why I listed out my steps. It wasn't working for me in anonymous mode, I couldn't log it.

skddc:
That error can certainly be connected, because the backend name is usually stored in localStorage, which is not available in this case. It *should* use memory, but it might not in some cases.

@galfert Maybe you can verify with the same app or some other app that uses widget login?

galfert:
Weird, I can't reproduce it in Safari private mode. But in normal Chrome I get a similar error:

>Uncaught TypeError: Cannot read property 'dropbox' of null @ remotestorage.js:6882 
activateBackend @ remotestorage.js:6882
initial @ remotestorage.js:6671
display @ remotestorage.js:6637
display @ remotestorage.js:6271
RemoteStorage.displayWidget @ remotestorage.js:6338
init @ app.js:21

I'm trying to find out what's going on.

galfert:
Found the culprit: It was due to the change from accessing localStorage via `localStorage['key']` to `localStorage.getItem('key')`. The former returns `undefined`, while the latter returns `null` for keys that don't exist.

That resulted in this line https://github.com/remotestorage/remotestorage.js/pull/940/files#diff-a18ce18f9db57ab2aeeb4b2e649342eaL198 not throwing an exception anymore and thus setting `apiKeys` to `null`.

It's fixed now, but I haven't gotten around to writing a testcase for this yet.

@silverbucket could you give this another try, please?

galfert:
I noticed that in private Safari I have to connect twice for the widget to show that I'm connected. Before I had only tested this in an app that was connecting without the widget.

I will create another issue for this, as it doesn't seem to be directly related to the actual exception this PR is fixing.

skddc:
If this PR is meant to fix rs.js in Safari private windows, then doesn't it belong here as well? It seems to me like it's directly related to localStorage not being available.

galfert:
I found the reason for having to connect twice, which is the following:

After the first connect the user gets redirected to the authorization of the remoteStorage provider. When getting redirected back to the app, the app gets the token, but at that time the `userAddress` and results from the Webfinger discovery are gone, because they couldn't be loaded from localStorage.
When doing a connect via the widget for a second time, the library sets the `userAddress` and the result from discovery. At that point, since the access token is already set from the first connect, the library has all required data and sets the `connected` state without another redirect.

I'm currently working on a fix for this that keeps track of the `userAddress` by adding it to the `redirectUri` using the `state` param and then extracting it again after the redirect.

skddc:
So we have to do a second Webfinger query?

galfert:
Yes, it's doing a second Webfinger query.

skddc:
Why not store all of the response in the state param then? Size limitations?

galfert:
I think I have a working solution now. I'm still writing tests, but maybe you can already give it a try and check if you see potential problems.

galfert:
> Why not store all of the response in the state param then? Size limitations?

It might become rather long and also hard to properly extract them all again. So I thought since it's such an edge case it's an acceptable tradeoff to do another Webfinger request instead.

skddc:
If there's no limitation, we could just base64-encode the whole JSON object. The URL is long either way, due to all the params, and not meant to be human-readable. Seems rather clunky to do the whole lookup again, after the authorization has been given, and also seems like a security concern to me. At least we'd have to review this from a security perspective and then maintain it with that in might over the years.

galfert:
I like the idea of base64-encoding the whole JSON object.

galfert:
FYI: I was trying to add all the discovery data as base64-encoded string to the redirect URL. But the problem is that the information is currently not accessible from where I need it (the Authorize object).

My current idea is to (after the Webfinger discovery) save all the data on the `remote`, as it's the information _about_ the remote and make it accessible from there. I guess something similar might need to be done for the other storage backends (Dropbox & Google Drive) as well, but I haven't looked at that yet.

skddc:
Save it how? If you only have in-memory storage, then that data is gone when you come back.

galfert:
"Save" it in memory on the `remote` object, so I can access it from the `authorize` object and add it as a parameter to the redirect URL. Then after the redirect get the data from that parameter and put everything back on the `remote`. That way no data is lost and there is no need for another Webfinger discovery after the redirect.

skddc:
Ah, got it. Thanks.

galfert:
Connecting to a remoteStorage backend is working without additional Webfinger request now.

Connecting to 3rd party backends doesn't work yet, though. First part of a fix for that is at #942.

skddc:
Very cool!

skddc:
@silverbucket Wanna try again?

silverbucket:
+1 works for me. Sorry I missed your message, must have been when I was in Norway.

gregkare:
@skddc and I have tested this on the current stable Safari (9.1.1). It works, +1

We found one caveat but this is expected because it is saving everything in memory: local RS data is lost when connecting your remote storage.

We have created [a Wiki page about private/incognito browsing mode](https://wiki.remotestorage.io/RemoteStorage.js:Private_browsing_mode) and documented Safari's behavior.

src/authorize.js
```diff
@@ -11,7 +11,27 @@
...
+        var stateValue = decodeURIComponent(kv[1]);
+        var encodedData = stateValue.substr(stateValue.indexOf('rsDiscovery='))
+                                    .split('&')[0]
+                                    .split('=')[1];
+
+        m['rsDiscovery'] = JSON.parse(atob(encodedData));
+
+        // remove rsDiscovery param
+        stateValue = stateValue.slice(stateValue.indexOf('rsDiscovery='),
+                                      encodedData.lenth + 12);
```
gregkare: `encodedData.lenth` should be `encodedData.length` (typo). It looks like this would remove all state data if there were keys after `rsDiscovery`

src/authorize.js
```diff
@@ -11,7 +11,27 @@
...
-      m[decodeURIComponent(kv[0])] = decodeURIComponent(kv[1]);
+
+      if (kv[0] === 'state' && kv[1].match(/rsDiscovery/)) {
+        // extract rsDiscovery data from the state param
+        var stateValue = decodeURIComponent(kv[1]);
+        var encodedData = stateValue.substr(stateValue.indexOf('rsDiscovery='))
+                                    .split('&')[0]
+                                    .split('=')[1];
+
+        m['rsDiscovery'] = JSON.parse(atob(encodedData));
```
gregkare: Could we find a more expressive name for that `m` variable?

src/remotestorage.js
```diff
@@ -763,8 +763,12 @@
     },
 
     localStorageAvailable: function () {
```
gregkare: Can't we reuse `RemoteStorage.LocalStorage._rs_supported` or even not have to check that feature since it was initialized already?

src/remotestorage.js
```diff
@@ -447,9 +447,9 @@
       this.backend = what;
       if (this.localStorageAvailable()) {
         if (what) {
-          localStorage['remotestorage:backend'] = what;
+          localStorage.setItem('remotestorage:backend', what);
```
gregkare: Question unrelated to these changes: if it doesn't set a backend at all (when local storage is not available), then how does it know which backend to use later on?

galfert: It does set a backend in line 447, even if localStorage is not available.

src/remotestorage.js
galfert: `RemoteStorage.LocalStorage` is only for the caching of nodes and not for storing configuration data. It might not even be included in some builds (e.g. the `nocache` build).

galfert:
Thanks for checking it out. I fixed/answered your comments.

src/remotestorage.js
skddc: Yes, but can't we just access the feature list in memory then? Isn't that initializing every all features on load? I.e. if the feature is loaded, then save.

galfert: But what about the `nocache` build? There you still want to save config in localStorage, even though that caching feature is not available.

src/util.js
```diff
@@ -401,8 +401,22 @@
       }
 
       return md5(str);
-    }
+    },
     /* jshint ignore:end */
+
+
+    localStorageAvailable: function() {
+      if (!('localStorage' in global)) { return false }
```
gregkare: This introduces an error when I'm building RS.js from this PR and use it in my favorite drinks:

```
ReferenceError: Can't find variable: global
```

gregkare:
Can everyone review/try this again? 8af183e broke it, I fixed it in 7efd35b

galfert:
@gregkare Thanks for fixing it. Too bad the tests didn't pick up on this, but `global` is available under node so it didn't fail there.

I found and fixed another problem with the `state` param extraction.

gregkare:
@galfert I've just tested this again in a private window in Safari (using a local build of my favorite drinks), this works.

If I reload the page after logging into my remote storage, I get the following error:

```
WireClient configure settings parameter should be an object
"configure@http://localhost:8000/remotestorage.js:5053:24
onFeaturesLoaded@http://localhost:8000/remotestorage.js:5902:41
http://localhost:8000/remotestorage.js:4632:22
forEach@[native code]
_emit@http://localhost:8000/remotestorage.js:4631:48
http://localhost:8000/remotestorage.js:3880:21
[native code]
http://localhost:8000/remotestorage.js:3941:21"
```

and the RS widget displays:

```
Sorry! An error occured.

undefined
```

skddc:
Perhaps we should extract that to a new nice-to-have issue, if private mode works at least without reloading now.

gregkare:
Agreed! This is already a good improvement

galfert:
Damn, there still seems to be a problem with the parameter extraction, so it's a problem with this PR. Sorry about that. I'm fixing it right now.

galfert:
Ok, fixed it. The `rsDiscovery` param wasn't removed properly from the URL after extracting it, so on reload remoteStorage tried to reuse it.

skddc:
It looks like the last few fixes are missing tests. Doesn't that mean it's likely to break unnoticed, next time someone touches this code?

gregkare:
@galfert This removes the param from the URL and fixes the error on reload

galfert:
> It looks like the last few fixes are missing tests. Doesn't that mean it's likely to break unnoticed, next time someone touches this code?

True. The last changes were done to a private function, so it's hard to test it directly. I'm currently trying to figure out how to test it via a public function that uses it's functionality.

skddc:
Cool.

galfert:
Wrote a test that covers all the recent fixes.

Found and fixed another potential issue along the way. Hooray for testing \o/

test/unit/authorize-suite.js
```diff
@@ -138,6 +138,44 @@ define(['requirejs', 'fs'], function(requirejs, fs, undefined) {
       },
 
       {
+        desc: "_rs_init extracts rsDiscovery data param for configuring the WireClient",
+        run: function(env, test) {
+          var storage = new RemoteStorage();
+
+          // mock the atob function, as it's not available in node
```
skddc: How is this handled when rs.js is run in node? Wouldn't it be polyfilled in the source?

galfert: That function is only used on the Browser redirect of the OAuth dance. So it's not used in node.

skddc: Right, thanks. Had that hunch too, after writing the comment. :)

galfert:
Can someone with OS X give this a try, please?

For your convenience I deployed two apps with the patched remoteStorage.js version.

My Favorite Drinks: https://mfd-safari.5apps.com
Webmarks: https://webmarks-safari.5apps.com

So you'd just need to test if everything works for you in Safari private browsing mode.

/cc @silverbucket, @jorinvo, @xMartin

gregkare:
I forgot to add an official +1 comment, so here it is: +1

skddc:
> So you'd just need to test if everything works for you in Safari private browsing mode.

... and [don't let Web Inspector troll you](https://wiki.remotestorage.io/RemoteStorage.js:Private_browsing_mode). ;)

xMartin:
Tried https://mfd-safari.5apps.com in current Safari on Mac and was able to load data, edit and have changes synced without errors.

galfert:
Thanks for testing, @xMartin!

galfert:
FYI: I'll merge manually, so I can squash some of the commits.

galfert:
Merged with 33c910d.

PR #942, opened 2016-07-08
(Re)initialize 3rd party backend when setting API key

### LINKED ISSUE #941: 3rd party backends don't get initialized properly on first app load
When loading the app for the first time, the app sets the API keys only after the remoteStorage initialization has already happened. But the backends only get initialized, when there is an API key set.

So the widget shows the backend's icon, but the class hasn't been initialized, which results in an exception when clicking the icon.

When loading the app for the second time and the API key is loaded from localStorage before the backend initialization, the backend object will be created properly.

galfert:
fixes #941

This fixes an issue where the backends wouldn't be initialized when the key was not already saved in localStorage before or the key is different from the one in localStorage.

### DIFF
--- a/src/remotestorage.js
+++ b/src/remotestorage.js
@@ -519,6 +519,13 @@
     setApiKeys: function (type, keys) {
       if (keys) {
         this.apiKeys[type] = keys;
+        if (type === 'dropbox' && (typeof this.dropbox === 'undefined' ||
+                                   this.dropbox.clientId !== keys.appKey)) {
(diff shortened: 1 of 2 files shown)


skddc:
lgtm +1

skddc:
Could someone else please review and either plus-one this or add comments/issues? Thanks!

cc @untitaker @silverbucket @Ragnis @michielbdejong @clochix

michielbdejong:
lgtm!

Issue #917, opened 2015-11-08
Replace bluebird with a simpler Promise polyfill


jorinvo:
Bluebird is huge.
I tried using a different promise polyfill and everything still works.
Seems like there are no special bluebird features used anywhere...
I safed 60kb _minified_ with this.

Also, the promise lib shouldn't be a hard dependency, because many users already include some polyfill anyway. 
Just tell the user that you use the global Promise and for old browsers the user needs a polyfill.

I used this one: https://github.com/jakearchibald/es6-promise

skddc:
Good idea.

We'll have to change some code however, as we currently use deferred objects quite a lot. Unfortunately,`Promise.defer` is now obsolete (and non-standard) and needs to be replaced in order to use vanilla Promises.

jorinvo:
@skddc I think RS already includes a version of defer: https://github.com/remotestorage/remotestorage.js/blob/master/src/remotestorage.js#L4

skddc:
Oh wow, totally missed that. No changes needed then. :)

silverbucket:
Yeah - defer was obsoleted as a build-in method, but it's easy for us to
just make our own. So that should be fine. +1 on removing bluebird.

gunar:
Good idea. Agreed.

jorinvo:
The only thing we  need to do is to decide if we want to bundle a Promise polyfill or tell the user to include one ...

galfert:
[localForage](https://github.com/mozilla/localForage) for example offers two separate builds. One with a bundled Promise polyfill (default) and one without. See (https://github.com/mozilla/localForage/tree/master/dist.

I think that's a good approach.

skddc:
... except we didn't notice that when using it, and when it shat all over our parade, it took an annoying debugging session to find out why. So the approach is demonstrably not that good after all; at least in terms of execution with that lib.

jorinvo:
I would tend to not include it. 
Promises [are coming](http://caniuse.com/#search=promise) to the browsers anyway and compared to the rest of setting up RS for an app, including a promise polyfill is pretty simple.
We can just add a one-liner to the instructions with a recommended polyfill lib.

skddc:
We could also at least throw a human-readable warning when `Promise` is not defined, and tell developers that way (maybe with a link to the docs).

galfert:
First step would be to get rid of the deprecated `deferred` usage.

skddc:
[...]

PR #946, opened 2016-08-05
[WIP] Remove Bluebird dependency

galfert:
connected to #917

* [x] Remove all deprecated `Promise.defer` calls
* [ ] Remove Bluebird
* [ ] Throw error when `Promise` is not defined (telling the developer to include a polyfill)

### DIFF
--- a/src/authorize.js
+++ b/src/authorize.js
@@ -101,39 +101,37 @@
    * Open new InAppBrowser window for OAuth in Cordova
    */
   RemoteStorage.Authorize.openWindow = function (url, redirectUri, options) {
-    var pending = Promise.defer();
-    var newWindow = global.open(url, '_blank', options);
+    return new Promise(function(resolve, reject) {
+      var newWindow = global.open(url, '_blank', options);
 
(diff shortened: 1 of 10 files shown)


skddc:
I'm not sure why we should drop `defer`. It's just a very small wrapper function that I think makes the code easier to read and write: https://github.com/remotestorage/remotestorage.js/blob/master/src/remotestorage.js#L4

At least doesn't really have anything to do with the issue at hand, which is removing Bluebird, not refactoring other code. No?

skddc:
@galfert Any updates/comments here?

galfert:
To be honest I missed @jorinvo's comment and thought we were using the `defer` method that's coming from Bluebird. I only realized this when I actually found the wrapper in `src/remotestorage.js` itself. But at that point I was almost finished with the refactoring.

I'm fine with just closing this whole PR, although I think in some of the cases it still might make sense to use the standard Promise way instead of `defer`.

skddc:
It'd just be easier to remove Bluebird without these changes I think.

galfert:
Fine with me.

Issue #945, opened 2016-08-01
Misuse of module.exports breaks remotestorage.js when a loader is used to load the script


dantman:
[L4151-L4156](https://github.com/remotestorage/remotestorage.js/blob/master/release/stable/remotestorage.js#L4151-L4156) use `module.exports` if present to expose `RemoteStorage` instead of defining `global.RemoteStorage`.

However soon after `RemoteStorage.util` attempts to use the global `RemoteStorage` directly.

This means the script will if you try to take advantage of the export/script loading features of Webpack, Browserify, etc...

skddc:
Thanks! Which version did you encounter the problem with?

dantman:
The version packaged into npm as `0.12.1`.

galfert:
I was able to reproduce the problem when trying out the current master's AMD build in [Litewrite](https://github.com/litewrite/litewrite).

I tried to figure out what changed between this version and the one that Litewrite is using at the moment. Then I noticed that Litewrite is actually using a patched version of `remoteStorage.amd.js` (e.g. with Bluebird removed). But I couldn't find the exact change that fixed this error.

@jorinvo Could you shed some light on what you did to fix this, please?

untitaker:
I'm hitting the same problems. It seems to me that remoteStorage could use `require` everywhere and just release its builds using webpack?

skddc:
Yes, I think ideally we switch the whole library to ES6 modules with Webpack or Browserify for packaging soon.

Also see #673, #786, #582, #501 for some related issues (that we could all solve in one go).

skddc:
@dantman We just released a new stable version, fixing the AMD build. Also, @untitaker wrote about how he's using it with Webpack in the wiki: https://wiki.remotestorage.io/RemoteStorage.js:Beginners'_Guide#Webpack

lesion:
I would love the whole library packed with webpack and with ES6 modules, and I could help in this task if you need !

I'm currently having issues because of this, trying to use the lib inside a chrome extension,
with chrome complaining about this `tv4` dependency (probably wrong packed 'cause I see they are using 
`this` for global context but `this` in chrome extension is not pointing to global `windows` object but to `undefined`)

skddc:
> I would love the whole library packed with webpack and with ES6 modules, and I could help in this task if you need !

That'd be fantastic!

[...]

PR #947, opened 2016-08-31
Fix AMD build

skddc:
See commit messages for details.

Connected to #945 
Connected to #917

### DIFF
--- a/build/builder.js
+++ b/build/builder.js
@@ -16,7 +16,7 @@ exports.build = function(components, selectedGroups, options) {
   if(options.amd) {
     exclude = components["amd-exclude"];
     console.error("AMD build: starting define() block.");
-    output += "define(['bluebird'], function(Promise) {\n";
+    output += "define([], function() {\n";
   }
 
   var files = [];
--- a/src/remotestorage.js
+++ b/src/remotestorage.js
@@ -878,9 +878,9 @@
    * Not available in no-cache builds.
    */
 
-  if ((typeof module === 'object') && (typeof module.exports !== undefined)){
+  global.RemoteStorage = RemoteStorage;
+
+  if ((typeof module === 'object') && (typeof module.exports !== 'undefined')){
     module.exports = RemoteStorage;
-  } else {
-    global.RemoteStorage = RemoteStorage;
   }
 })(typeof(window) !== 'undefined' ? window : global);


untitaker:
As a JS beginner I'm still confused how to use remoteStorage with npm and webpack.

skddc:
Webpack confuses everybody, no matter their JS skill level. ;)

If you use it from npm, you'd just link the `release/...` build in whatever front-end builder you're using (if you're not using it in node.js that is). Or even just have a task that copies it from there to a local folder at a certain point. We're using `release/stable` as path for stable buids now (instead of a version number/name), so you can more easily upgrade via npm.

galfert:
LGTM. +1

untitaker:
I resorted to copying the resulting `.amd.js` into my repo. I also had to patch the file like this:

```diff
diff --git a/home/untitaker/projects/remotestorage/remotestorage.js/remotestorage.amd.js b/lib/remotestorage.amd.js
index f999fa2..5da0933 100644
--- a/home/untitaker/projects/remotestorage/remotestorage.js/remotestorage.amd.js
+++ b/lib/remotestorage.amd.js
@@ -2131,10 +2131,6 @@ define([], function() {
  *
  */
 
-if (typeof XMLHttpRequest === 'undefined') {
-  XMLHttpRequest = require('xmlhttprequest').XMLHttpRequest;
-}
-
 (function (undefined) {
 
   // URI to property name map

```

Otherwise webpack would look for a xmlhttprequest dependency.

Just installing from npm wouldn't work because that would pull in a `xmlhttprequest` library that itself requires a `fs` module.

galfert:
What? That is only supposed to load that library in Node.

untitaker:
Yes, and apparently webpack can't understand the conditional and thinks that the sourcecode requires `xmlhttprequest`. Which is why I removed the entire thing.

skddc:
> What? That is only supposed to load that library in Node.

Perhaps we should use https://github.com/remotestorage/remotestorage.js/blob/master/src/env.js#L16-L18 then. But not sure if Webpack knows about `window` there.

untitaker:
This `webpack.config.js` fixed it for me:


    module.exports = {
        externals: [
            {xmlhttprequest: '{XMLHttpRequest:XMLHttpRequest}'}
        ]
    };

skddc:
Great!

@untitaker @jorinvo Perhaps you could create a little wiki page about using Webpack and/or update https://github.com/remotestorage/remotestorage.js/blob/master/doc/amd.md?

@jorinvo It seems this PR would be enough then, not withstanding better instructions for new users.

skddc:
@jorinvo Could you please take a minute to confirm this fix? Thanks!

jorinvo:
@skddc I don't actively follow the development of remotestorage and I can't actually test this change, but this looks like the right thing to do. 
Removing bluebird is more a performance question I think.
And having a global plus a module.export version looks right. 
Not sure if there are other missing globals interally. Would need to run it to see if it's working. 
Has been really long ago that I was looking into this.

theWebalyst:
Noob warning... but I believe this now works out of the box.

I merged bugfix/945-amd_build_issues with my SAFEnetwork code, rebuilt remotestorage.amd.js, linked to that from litewrite/lib, then ran ```npm run build``` and litewrite appears to work. I don't know its features, but reckon if it starts and I can create a note we're probably good. I have not tried connecting it to any storage tho, so only a superficial test.

I can do some more tests if you want to give me some pointers as to what might be the best things to do, but this has removed the need for me to hack the remotestorage.amd.js in order to get it to function at all.

Hope that helps a bit!

skddc:
> Would need to run it to see if it's working.
> Has been really long ago that I was looking into this.

I could run this build with Litewrite. So it should be easy for you as well to do the same.

skddc:
> I can do some more tests if you want to give me some pointers as to what might be the best things to do, but this has removed the need for me to hack the remotestorage.amd.js in order to get it to function at all.
> 
> Hope that helps a bit!

Yes, it does! The problems were based on loading the libs, so if it works at all, that's confirming the fix. Thank you.

skddc:
Taking @theWebalyst's report as second +1 and merging. Thanks, everyone!

untitaker:
Here's the usage I had success with using Webpack: https://wiki.remotestorage.io/RemoteStorage.js:Beginners%27_Guide#Get_remoteStorage.js

skddc:
Great, thanks!

untitaker:
That's just what I came up with. Perhaps somebody else has found a simpler way (especially regarding the XmlHttpRequest shim).

PR #948, opened 2016-09-06
Fix discovery errors (and others) not being thrown/shown

### LINKED ISSUE #937: No discovery error thrown/shown on Webfinger 404 responses
Someone stumbled upon this issue in Litewrite:

https://github.com/litewrite/litewrite/issues/268

The Gist is: 5apps.com is responding with a 404 to Webfinger requests for non-existing users and remoteStorage.js thinks that's not a good enough response and continues its discovery dance by falling back to http, which in turn fails because we don't serve http with CORS headers, and it would only be a redirect to https anyway.

Instead of throwing a discovery error and showing it in the UI, this just triggers console errors and breaks the app/widget however.

Now, the first question of course is, if 404 is a valid response when asking for user information of a user that doesn't exist. Unfortunately the Webfinger spec is so incredibly vague in that regard, that it's basically useless. The only part I found was:

> Likewise, if a server has no information to provide, it might return a JRD with an empty "links" array or no "links" array.

It *might*. Not MAY, SHOULD or MUST, but it just *might*. Whatever else it *might* do instead is left to the reader's imagination it seems.

Does anyone have insights or opinions on this topic? Should we treat it as a bug and accept 404s, should 5apps change it to the *might* quoted above, should we spec the case in the RS draft, ...?

skddc:
After updating webfinger.js with the version including @silverbucket's fixes for Webfinger 404s not executing any callback functions, I found that no error messages were shown in the widget anymore. Not exactly sure when that broke, but my small refactoring of how they get passed around in the code fixes the problem.

However, in an actual app, I'm seeing weird behaviour when Webfinger fallbacks are being tried (hence the WIP title). There are two different discovery errors being thrown, and the first one claims that it couldn't connect to the storage server, while the second one (a good multiple seconds later), correctly states that it couldn't find storage information on the server. Both are updating the widget, so the first messages is being replaced with the second one after a couple of seconds. Here's what it looks like in my console:

screenshot from 2016-09-06 13-42-20

I did see that webfinger.js is throwing multiple errors now, but they were all the same afaics, so it doesn't explain to me why we get two different ones in rs.js. I'll try to find out what's happening exactly and report back.

connected to #937 
fixes #937

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
   "dependencies": {
     "bluebird": "^3.0.5",
     "tv4": "^1.2.7",
-    "webfinger.js": "^2.2.1",
+    "webfinger.js": "^2.4.2",
     "xhr2": "^0.1.3"
   }
 }
--- a/src/discover.js
+++ b/src/discover.js
@@ -45,7 +45,7 @@
 
     webFinger.lookup(userAddress, function (err, response) {
       if (err) {
-        return pending.reject(err.message);
+        return pending.reject(err);
       } else if ((typeof response.idx.links.remotestorage !== 'object') ||
                  (typeof response.idx.links.remotestorage.length !== 'number') ||
                  (response.idx.links.remotestorage.length <= 0)) {
--- a/src/remotestorage.js
+++ b/src/remotestorage.js
@@ -364,7 +364,7 @@
       this._emit('connecting');
 
       var discoveryTimeout = setTimeout(function () {
-        this._emit('error', new RemoteStorage.DiscoveryError("No storage information found at that user address."));
+        this._emit('error', new RemoteStorage.DiscoveryError("No storage information found for this user address."));
       }.bind(this), RemoteStorage.config.discoveryTimeout);
 
       RemoteStorage.Discover(userAddress).then(function (info) {
@@ -396,7 +396,8 @@
           }
         }
       }.bind(this), function(err) {
-        this._emit('error', new RemoteStorage.DiscoveryError("Failed to contact storage server."));
+        clearTimeout(discoveryTimeout);
+        this._emit('error', new RemoteStorage.DiscoveryError("No storage information found for this user address."));
       }.bind(this));
     },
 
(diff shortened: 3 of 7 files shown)


silverbucket:
I had a look around, because those error messages did not look familiar to me, and found that the first case is the actual rejected promise, however the error message is not being used from the actual error, it's just using a generic error message statically written: https://github.com/remotestorage/remotestorage.js/blob/master/src/remotestorage.js#L399

The second is due to a setTimeout that is used if there is no satisfactory response. (Note: webfinger.js has a timeout function itself, and the exact timeout can be specified during object initialization so this is probably not needed)
https://github.com/remotestorage/remotestorage.js/blob/master/src/remotestorage.js#L366

silverbucket:
If you want to be able to write your own error messages and not use the ones given from webfinger.js, you could return the error object with the rejected promise, instead of just the message property of the error object, here: https://github.com/remotestorage/remotestorage.js/blob/master/src/discover.js#L48

Then decide what kind of message you want to write based on the status property (return code).

skddc:
Timeout was exactly what I was expecting to find there somewhere, based on the console logs. :)

Thanks for checking it out! I'll rewrite the discover.js code a bit then, based on your findings and suggestions.

skddc:
> Note: webfinger.js has a timeout function itself, and the exact timeout can be specified during object initialization so this is probably not needed

I just had a look at the webfinger.js source, but I couldn't find the `request_timeout` other than in the README and config code. It seems to me like it's not actually implemented.

skddc:
Due to timeout potentially missing as an error source otherwise, I left our own timeout functionality in the code. But I fixed it to also clear that timeout, when Webfinger.js throws an error before it is reached.

I also changed the different generic error messages to be the same single one. I think there's not much point in explaining in detail what went wrong under the hood anyway.

So this is ready for review/merge now!

silverbucket:
Wow, I never realized the timeout feature wasn't implemented in webfinger.js :) Just implemented it now.

reviewed this code, +1

galfert:
LGTM. +1

skddc:
Thanks.

Issue #949, opened 2016-09-22
Refactor the library to actually not need global variables and work properly with CommonJS/AMD/ES6 module systems.


lesion:
Like @skddc wrote in d4dd748, I agree that a refactoring is needed to simplify the usage of remoteStorage.
### Now:
1. Download remotestorage.js
2. Download a data module
3. Choose a build (AMD / nocache / min ...)
4. Include in your html

``` html
<script src='remotestorage.js'></script>
<script src='remotestorage-bookmarks.js'></script>
```
### Now with webpack / browserify / bower / npm

This is currently not working or at least I'm not able to make it works!
1. `npm install remotestoragejs --save-dev`
2. `npm install remotestorage/modules --save-dev` // fetch from github not from npm registry

``` js
require('remotestoragejs/release/stable/remotestorage.amd.js');
// window.RemoteStorage is now available
require('remotestoragejs-modules/bookmarks');
// dependencies issues & RemoteStorage in global scope !
```
### What do I expect
1. `npm install remotestoragejs --save-dev`
2. `npm install remotestoragejs-modules --save-dev`

``` js
var RemoteStorage = require('remotestoragejs')
var Bookmarks = require('remotestoragejs-modules/bookmarks')

// ES6 
// import RemoteStorage from 'remotestoragejs'
// import { bookmarks } from 'remotestoragejs-modules'
```

I think this will result in a cleaner code, without polluting the global namespace,
with declared dependencies between different part of the library.
let me know what do you think, to me porting the lib to 2016 and makes it
usable with modern tools will attract more people to working with it.
I think this is not a lot of work and if you agree with me I'm starting to do it.

skddc:
By the way, I think we should also remove the no-cache build with this move. At the moment, disabling caching in the normal build will still cache with the SEEN strategy I think, so that might require an API change or addition (or we can load based on name or sth).

lesion:
Nice for me to have a npm package for each single modules.

I've started to work on removing global RemoteStorage var and package the lib
with webpack in UMD (so everything could import it) on [feature/949-noglobal](https://github.com/remotestorage/remotestorage.js/compare/feature/949-noglobal) branch.

It's pretty much working right now with my setup but only in browser,
not in node because there's the XMLHttpRequest dependency from webfinder.js !
To fix this I think we can tell webpack to export a separate node package (dirty) or to 
convince @silverbucket to stop requiring `xmlhttprequest` and use an isomorphic
[...]

PR #951, opened 2016-10-25
Feature/949 noglobal

lesion:
refs #949
connected to #949

Do NOT merge into master, this is a work in progress and will break everything!


skddc:
Just a side note, seeing that builds are put in `/build` there: we keep release builds in `/release` (`/stable` for versioned ones, and `/head` for builds from master).

lesion:
ok for the builds.
an update on this branch status...
I'm currently having an issue at making a build that works in every environment with proposed plugin schema; this is now implemented in module-bookmarks, widget and the library itself. made all of them interact well with <script>, in es5 environment, with babel and also in node, it's not as easy as expected for three main reasons: first, changing the prototype it's not the cleanest way to do this kind of things (this is valid for the widget), then because there are dependency not compatible with every environment (I'm currently struggling with the `global` keyword and how webpack handle it). But mainly is the lack of documentation of webpack ([this](https://webpack.github.io/docs/library-and-externals.html) is the only thing I've found on this topic).
This is taking more time than expected as I can work on it only in spare time, but I'm keep working on this. hope to find a clean solution soon.

lesion:
ps. another thing I've found, webpack does not behave well with symlinks, so using the useful `npm link` command while doing this kind of things its unrecommended

skddc:
Thanks for the update, and thanks for continuing work on it, even though it's considerably more than you expected!

Unfortunately, `global` is only defined in node.js, and acts like a normal (undefined) variable anywhere else. We're just natively supporting node.js with that right now (the node package will actually load things from `src`, not from a release build).

So I think if we support node.js via an actual build, we can actually remove that logic from the source files, because we don't need it anymore. We'll just have to be careful about it potentially being used as a feature support flag here and there (i.e. "if node.js, then foo" being written as `if (typeof global !== undefined) { }`").

lesion:
my mistake! it was working as expected. The issue was in my [rs-test repo](https://github.com/lesion/rs-test), I was [transpiling ES6 code](https://github.com/lesion/rs-test/commit/acc4517e4f27fa2554f38e172d7e8ccaf8ce3fce#diff-a6d7b8359d6577acdecd3d19c459c4fbR19) twice (and this was causing a [known](https://github.com/webpack/webpack/issues/1963) issue with `global`).
Each environment in https://github.com/lesion/rs-test is now working as expected. please PR if you wanna add new environment to test (like `browserify`, `rollup`, `webpack2` etc..)

I'm now working with `unit tests` and I'm thinking to start reversing the dependency tree in source code (as [explained](https://github.com/remotestorage/remotestorage.js/issues/949#issuecomment-255224955)). Please check if the last commit is ok for you (is an example with `access` module and it passes all tests of the module itself)

skddc:
> Please check if the last commit is ok for you (is an example with access module and it passes all tests of the module itself)

Yes, that looks good to me.

lesion:
a little question here, why RemoteStorage is a constructor ? I mean, there's an use case in using 2 instances of RemoteStorage in the same page / project / app ? If not, why it needs to be a constructor? I think actually you can't use more then one instance (correct me if I'm wrong).

I'm asking this because I see a lot binding to `this`, a lot of messy with prototype of RemoteStorage that will be not needed without the necessity of using more instances of the same constructor (and I think it's the case, also, I don't think you can actually use more than one instance, correct me if I'm wrong)

skddc:
> I mean, there's an use case in using 2 instances of RemoteStorage in the same page / project / app ?

Yes, that's the idea and also concrete plan for the near future. It should be possible to connect multiple accounts in one app.

lesion:
Sorry for the noise ! I need to study something about git I suppose. I'd `pull --rebase` from master to merge your last changes here, didn't knew that this would cause all these commits again..

An update from this branch: 
I'm pretty close to ask you to try it ....

result from `npm run test`
``` scaffolding report  0 failed,  1492 passed, 0 skipped, 772 total.
       test report  2 failed,  358 passed, 0 skipped, 360 total.
```
Before trying it, I've a last thing to solve in how internal features are initialized.
hope to finish it this week.

skddc:
youre-awesome

lesion:
\o/
ok this is getting interesting (:
While finish to clean up the mess I've done here and there, I've started to write some documentation on changes, you can find it [here](https://github.com/remotestorage/remotestorage.js/blob/feature/949-noglobal/doc/noglobal.md). 
feel free to explore and to ask.

skddc:
Just a little heads up: I had to rebase the nowidget branch with conflicts and force-pushed it afterwards, so there are likely some conflicts on the way before merging this. Nothing impossible though.

I was wondering about this however: if Webpack supports all kinds of import mechanisms, would it be possible to replace the `require` statements with ES6 `import` ones, so that eventually we could just drop the source in an app and use it as is? (Maybe it's a stupid question, I'm not too familiar with either the module systems or Webpack.)

lesion:
> Just a little heads up: I had to rebase the nowidget branch with conflicts and force-pushed it afterwards, so there are likely some conflicts on the way before merging this. Nothing impossible though.

I see, I wanted to try the new `Resolve conflict` feature of github but it's not properly working right now, I'll try so solve the conflicts these days...

> I was wondering about this just now: if Webpack supports all kinds of import mechanisms, would it be possible to replace the require statements with ES6 import ones, so that eventually we could just drop the source in an app and use it as is? (Maybe it's a stupid question, I'm not too familiar with either the module systems or Webpack.)

Yes, this could totally work, but we'll have an issue with tests. As ES6 (`import` statement is ES6) is transpiled by webpack (via babel) the library will be in plain ES5, but as unit tests are requiring module directly this will be a syntax error (native support in node for `import` is [an hot potato](http://stackoverflow.com/questions/37132031/nodejs-plans-to-support-import-export-es6-es2015-modules)).
I have the same problem trying to use some new ES6 construct (like arrow functions) and tests is passing only with a not so old node's version (>=4)

I'm not so familiar with unit testing but I think a refactoring as to be done there.

skddc:
> Yes, this could totally work, but we'll have an issue with tests. As ES6 (import statement is ES6) is transpiled by webpack (via babel) the library will be in plain ES5, but as unit tests are requiring module directly this will be a syntax error (native support in node for import is an hot potato).

Ah ok. So it would also not work in node.js in general for now.

> I have the same problem trying to use some new ES6 construct (like arrow functions) and tests is passing only with a not so old node's version (>=4)

That's totally fine imo. We can just require a modern node.js version in `package.json`. It's 6+ already, so shouldn't be a problem to require 4+.

skddc:
> I'm not so familiar with unit testing but I think a refactoring as to be done there.

Let's just do it step by step and get it working with `require` for now. It's already such a big step forward and kind of a huge PR.

lesion:
Ok, I think I've done with this, there are two small fix left to remove `defer` Promise use, but I need some help with these: 
https://github.com/remotestorage/remotestorage.js/blob/4a21c37b3333291b9abd74759e87f1eff7bbe991/src/cachinglayer.js#L298
https://github.com/remotestorage/remotestorage.js/blob/4a21c37b3333291b9abd74759e87f1eff7bbe991/src/remotestorage.js#L532


I know, I made a lot of decisions behind without discussing it and I'm sorry for that, please feel free to ask and take a look at the little document I write on changes: https://github.com/remotestorage/remotestorage.js/blob/4a21c37b3333291b9abd74759e87f1eff7bbe991/doc/noglobal.md

happy hacking in 2017 !

skddc:
@lesion I just added this branch to an Ember app (also with [its own tiny module](https://github.com/67P/remotestorage-module-kosmos), based on your bookmarks module) and it works very well so far. Great job!

The only bug I stumbled upon, but I haven't investigated it much so far, is that when Trying to call `remove()` on a BaseClient (while not connected; haven't added widget yet), it throws an exception somewhere in `doUpdateNodes`:

```
ReferenceError: assignment to undeclared variable previous
```

Did you have any issues with `remove` in the bookmarks module?

lesion:
I've not using `remove()` anywhere by now so I cannot really help you with this. For sure this branch is not actually ready to be merged, has to be deep tested before. I'm having some issue too working with my extension, I will try to check it out next week.

btw:  https://chrome.google.com/webstore/detail/memm/nmmkepllfhgjonodmhgaoaehnknoadfo
not published to the world yet as it's undocumented by now (it is using a bookmarks module in the same namespace and format as your original bookmark module, so it is completely compatible with https://webmarks.5apps.com)

skddc:
> I've not using remove() anywhere by now so I cannot really help you with this. For sure this branch is not actually ready to be merged, has to be deep tested before. I'm having some issue too working with my extension, I will try to check it out next week.

Ok, cool. I'll keep building with this branch in order to test it more. If I see something I can help fix or at least investigate further, I'll also try to do that.

> btw: https://chrome.google.com/webstore/detail/memm/nmmkepllfhgjonodmhgaoaehnknoadfo
> not published to the world yet as it's undocumented by now (it is using a bookmarks module in the same namespace and format as your original bookmark module, so it is completely compatible with https://webmarks.5apps.com)

Great! Will test that too.

skddc:
I just ran into two more things:

1. `Promise` isn't required from bluebird anymore, which makes the library fail in non-ES6 environments (like e.g. the current PhantomJS stable version). We wanted to remove bluebird in favor of telling people to use a polyfill anyway however, so I think there's not really anything to fix there. We have [a seperate issue](https://github.com/remotestorage/remotestorage.js/issues/917) open for it, where we can remove the dependency.

2. This might sound a bit special, but when using `browserify` with the rs.js npm module (from a app's codebase) and its release build, then `xmlhttprequest` can be missing from the dependencies. I only had this issue on Travis CI, and not locally, so it's probably a default option flag for the `npm install` on Travis. I think the fix would be to move it from `devDependencies` to `dependencies`, but I'm not sure. I could fix it for our app by just adding `xmlhttprequest` to the app's own `package.json`, so that Browserify can find it.

lesion:
Your issue with `remove()` was due to an undeclared var. Last commit fixed it.

> Promise isn't required from bluebird it seems, which makes the library fail in non-ES6 environments (like e.g. the current PhantomJS stable version). We wanted to remove bluebird in favor of telling people to use a polyfill anyway however, so I think there's not really anything to fix there. We have a seperate issue open for it, where we can remove the dependency.

Yep, I saw that issue and I agree not exporting a Promise polyfill (we need to document this of course).

> This might sound a bit special, but when using browserify with the npm module and its release build (from the app's codebase), then xmlhttprequest can be missing from the dependencies. I only had this issue on Travis CI, and not locally, so it's probably a default option flag for the npm install on Travis. I think the fix would be to move it from devDependencies to dependencies, but I'm not sure. I could fix it for our app by just adding xmlhttprequest to the app's own package.json, so that Browserify can find it.

yep, that dependency it was a bit annoying [from the beginning](https://github.com/remotestorage/remotestorage.js/issues/949#issuecomment-249328589).
In webpack enviroment you have to do [this trick](https://wiki.remotestorage.io/RemoteStorage.js:Beginners%27_Guide#Webpack), probably there's something similar in browserify.

skddc:
> Your issue with remove() was due to an undeclared var. Last commit fixed it.

Yay, thanks!

> yep, that dependency it was a bit annoying from the beginning. In webpack enviroment you have to do this trick, probably there's something similar in browserify.

I think going the Fetch-with-polyfilll route would be the best solution. /cc @silverbucket

silverbucket:
Hey guys, I'm still not sure I understand the issue with using `xmlhttprequest`. Perhaps it's just my ignorance of webpack, but can someone explain why it's a problem to use XHR requests in `webfinger.js`? I looked at it before and not only would I have to re-write a lot of the code to use the `fetch` API, but users of the library would then have to include a fetch polyfill if they want to be sure it will function on all browsers, which increases the size of the library. The reason I used the xmlhttprequest polyfil is because it's in the node environment where size isn't an issue.

lesion:
> Hey guys, I'm still not sure I understand the issue with using xmlhttprequest. Perhaps it's just my ignorance of the new setup, but can someone explain why it's a problem to use XHR requests in webfinger.js? I looked at it before and not only would I have to re-write a lot of the code to use the fetch API, but users of the library would then have to include a fetch polyfill if they want to be sure it will function on all browsers, which increases the size of the library.
> 
> The reason I used the xmlhttprequest polyfil is because it's in the node environment where size isn't an issue.

I get your point and you're right.
I don't want you to rewrite the lib with `fetch` (also because we would also need to rewrite some of remotestorage itself), but I want to find a way to make both environments (node & browser) the same to devs that wants to use remotestorage (shipping only a file, or using `npm install remotestorage` not `npm install remotestorage-node` and `npm install remotestorage-client` ).
if put `xmlhttprequest` as dependency in package.json could solve this issue for browserify, it's super ok to me :)

skddc:
I just saw that we already bundle `xhr2` in `dependencies`, so maybe we can unify that? Otherwise we have two dependencies for exactly the same thing.

skddc:
@silverbucket @lesion Any idea about my last question?

silverbucket:
Yeah I remember looking into both of these and making a decision to go with `xmlhttprequest`, though I don't remember all the reasons why. Looking at the projects now, it's clear `xmlhttprequest` is more popular, though why specifically is not obvious. However `xmlhttprequest` hasn't been touched for 2 years now, whereas `xhr2` was last updated 3 months ago. Also, `xmlhttprequest` only supports XHR v1 and while there were plans for v2, they never seemed to materialize. 

If I was evaluating now, I guess I'd go with `xhr2`. It should be more or less a drop-in replacement, with just a change in how it's initialized, so I can try to do this soon.

skddc:
> If I was evaluating now, I guess I'd go with xhr2. It should be more or less a drop-in replacement, with just a change in how it's initialized, so I can try to do this soon.

That'd be great, as it would solve this part then.

silverbucket:
@skddc v2.6.3 now uses `xhr2`

silverbucket:
* v2.6.4 - forgot to remove the old dependency

galfert:
I posted some ideas about possible API changes/improvements that I would like to get your opinions on: https://community.remotestorage.io/t/ideas-about-api-changes-additions/380

lesion:
green light :)

skddc:
I just wanted to merge, but there's an issue that prevented it:

The stable release build needs to stay the actual stable release build, not the alpha. Even the HEAD build needs to stay the current HEAD build in master for now. Those are actually used from e.g. Bower. We can change the HEAD build once we've publicy documented the new usage, but the stable build needs to stay in there until at least RC or even the final 1.0 release.

skddc:
I'm almost done fixing that locally, and I'll merge locally too.

skddc:
Done!

great-success

skddc:
Many thanks again @lesion, you're the undeniable hero of this chapter!

galfert:
Wohoo. Great work @lesion.

high5

lesion:
:tada: thanks!

Issue #923, opened 2016-02-19
Don't throw error for failed network/sync requests


skddc:
It's a normal situation that network requests fail, especially for an offline-first library used on mobiles. So I think we shouldn't throw actual exceptions when it happens, but rather only log a warning and maybe have an event for it that devs can use for e.g. analytics code.

skddc:
Anyone has opinions on this?

/cc @michielbdejong @silverbucket @untitaker @galfert

untitaker:
I don't have any opinion since the user doesn't look at console.log, though it's probably for the better when app devs are actually showing those errors to the user.

gunar:
we should dispatch an event so the app can inform the user if it's
online/syncd or not. devs could use the same event for debugging.

skddc:
Great, so let's turn it from an exception into an event.

Related: it'd be great to have a way to log all events from rs.js as a developer. Makes understanding the library's behaviour that much easier. @galfert and I once talked about building a browser extension, similar to Ember Inspector for that, which could also do other things, but we didn't ever do it. Maybe it's actually enough or even better to do it directly in the library, similar to [`enableLog()`](https://remotestorage.io/doc/code/files/remotestorage-js.html#RemoteStorage.enableLog) or even bound to that.

skddc:
I just hit this exact spot with my new-widget work. The current widget goes into offline mode when a request fails and comes back online when one succeeds. `remoteStorage.remote.online` is set accordingly. But that's actually where the correct event is missing I think. Instead of a "network request failed" event, I think we need a general "went offline/online" event.

Not sure about the naming/API, though. How about this:

```js
remoteStorage.on('network-offline', () => handleOffline);
remoteStorage.on('network-online', () => handleOnline);
```

Another option:

```js
remoteStorage.on('network', (evt) => {
    switch (evt) {
        case 'offline':
            handleOffline();
            break;
        case 'online':
            handleOnline();
            break;
    }
});
```

I like number 1 better, and we do already have 'connected' and 'disconnected' for the storage itself. What do you think?

skddc:
One thing we could add as additional event for "network" would be "synced", which we are also lacking as a specific event at the moment. Or it could be its own "sync" event with "started", "done" and so on as arguments or suffixes.

skddc:
> Or it could be its own "sync" event with "started", "done" and so on as arguments or suffixes.

There's actually an event for this in the `Sync` class, but it's not documented and seems to be private API so far.

[...]

PR #952, opened 2016-11-07
Add network-offline and network-online events

skddc:
This adds events for the network going offline (on every failed request) and going back online (on every succeeding successful request). It also moves the setting of remote.online on failure to the right place (same one that success is using).

I'm using this for the new connect widget, but it's also extremely useful for custom widgets/UI and app developers.

refs #923
connected to #923

### DIFF
--- a/src/remotestorage.js
+++ b/src/remotestorage.js
@@ -167,6 +167,18 @@
      *
      * Fired when a wire request completes
      **/
+    /**
+     * Event: network-offline
+     *
+     * Fired once when a wire request fails for the first time, and
+     * `remote.online` is set to false
+     **/
+    /**
+     * Event: network-online
+     *
+     * Fired once when a wire request succeeds for the first time after a
+     * failed one, and `remote.online` is set back to true
+     **/
 
     // Initial configuration property settings.
     if (typeof cfg === 'object') {
--- a/src/remotestorage.js
+++ b/src/remotestorage.js
@@ -176,8 +176,9 @@
 
     RemoteStorage.eventHandling(
       this, 'ready', 'connected', 'disconnected', 'not-connected', 'conflict',
-            'error', 'features-loaded', 'connecting', 'authing', 'wire-busy',
-            'wire-done', 'sync-interval-change'
+            'error', 'features-loaded', 'connecting', 'authing',
+            'sync-interval-change','wire-busy', 'wire-done',
+            'network-offline', 'network-online'
     );
 
     // pending get/put/delete calls.
--- a/src/remotestorage.js
+++ b/src/remotestorage.js
@@ -189,7 +189,7 @@
     RemoteStorage.eventHandling(
       this, 'ready', 'connected', 'disconnected', 'not-connected', 'conflict',
             'error', 'features-loaded', 'connecting', 'authing',
-            'sync-interval-change','wire-busy', 'wire-done',
+            'sync-interval-change', 'wire-busy', 'wire-done',
             'network-offline', 'network-online'
     );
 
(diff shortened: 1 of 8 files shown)


skddc:
* [ ] Add docs

src/remotestorage.js
```diff
@@ -176,8 +176,9 @@
 
     RemoteStorage.eventHandling(
       this, 'ready', 'connected', 'disconnected', 'not-connected', 'conflict',
-            'error', 'features-loaded', 'connecting', 'authing', 'wire-busy',
-            'wire-done', 'sync-interval-change'
+            'error', 'features-loaded', 'connecting', 'authing',
+            'sync-interval-change','wire-busy', 'wire-done',
```
galfert: Missing space before `'wire-busy'`.

skddc:
We'd like to release today, but are missing a second +1. Can someone check it out real quick? Code review would be enough, because @galfert and I will be testing it manually with the release candidate again.

/cc @michielbdejong @silverbucket @jorinvo @lesion @Ragnis @untitaker

untitaker:
This appears to make sense but I'm completely unfamiliar with the codebase.

gregkare:
+1

skddc:
Re-adding WIP, because this is not tested with Dropbox and Google Drive.

galfert:
I tested it with Dropbox. The offline mode seems to be broken in general. The `network-offline` event is not sent, and `remoteStorage.remote.online` is still `true`. There are no new attempted network requests afterwards. The widget shows the storage to be offline though. And after going online again the storage stays offline.

skddc:
@galfert Yes, we need to implement the events for the various backends as I suspected. Dropbox network events are working now (but no tests so far).

The Dropbox backend is definitely not issuing any more requests after the first one fails. I opened #953 about it.

skddc:
Added network events for Google Drive as well (also without tests for now).

Sync works fine there when going offline btw, so we can compare code when checking out the Dropbox issue.

galfert:
Tested both Dropbox and Google Drive again.

With Dropbox I do see the `network-offline` now.

With Google Drive I can see both `network-online` and `network-offline` events. But for me the widget still said "Offline" after coming back online (and getting the `network-online` event).

skddc:
> With Dropbox I do see the network-offline now.

So when you manually restart sync after it goes offline (due to #953), it doesn't give you the online event?

> But for me the widget still said "Offline" after coming back online (and getting the network-online event).

The widget doesn't say "offline", but "synced x ago". Do you mean the class attribute and icon color? If you see the event in the console, then it's highly unlikely the class wouldn't change, as that's being done in the same function. Could you please check why that code isn't called a few lines after in your browser?

galfert:
After manually restarting the sync I **do** see the `network-online` event, but the widget still says "Offline":

dropbox_offline

I'll check if/why the widget code doesn't get called.

skddc:
Oh. Now I see. I'm developing this with the new widget. I'm not fixing anything with the old widget anymore, because it's wasted time. The new one is almost finished.

galfert:
I found the reason for this is that both the Dropbox and Google Drive client are not sending the `wire-busy` and `wire-done` events. Even though the new widget will not be affected by this, I would still consider this a bug. Especially since both clients actually [announce to be sending those events](https://github.com/remotestorage/remotestorage.js/blob/6a397f393ebb9b27d869640f6bc474d7dc8f5fd1/src/dropbox.js#L196).

I have a fix for this already, but I'm currently working on the tests. But as you said, this is not really an issue of this branch itself.

galfert:
I added some specs for these network events to the Dropbox and Google Drive suites.

In connection with #956 and #958 these 3rd party clients work now, also with the current widget.

Overall I think this is good to merge now.

PR #954, opened 2016-11-13
Document `remoteStorage.startSync()`

skddc:
Stumbled upon this function by chance, while reading source code in the sync class. Looks like this is the public API for starting a sync explicitly.


galfert:
LGTM. +1

PR #956, opened 2016-11-18
Keep Dropbox client trying to sync when offline

### LINKED ISSUE #953: Sync requests stop when going offline
Reproduced by both @galfert and me: when you go offline with the Dropbox backend connected, it just stops issuing sync requests altogether.

galfert:
fixes #953 

Took me a while to find the cause, but this fixes the problem of the Dropbox client stopping all sync requests when going offline. When going back online, the sync requests start working again.

In conjunction with #952 it also sends the `network-online` event.

I haven't found a good way to write a test for this yet.

Also, same as [I mentioned for the Google Drive client](https://github.com/remotestorage/remotestorage.js/pull/952#issuecomment-261538725), after coming back online the widget still says "Offline". So far I haven't found the cause for that.

### DIFF
--- a/src/dropbox.js
+++ b/src/dropbox.js
@@ -698,8 +698,8 @@
       }, function (err) {
         this.rs.log('fetchDeltas', err);
         this.rs._emit('error', new RemoteStorage.SyncError('fetchDeltas failed.' + err));
-        promise.reject(err);
-      }).then(function () {
+        return Promise.resolve(args);
+      }.bind(this)).then(function () {
         if (self._revCache) {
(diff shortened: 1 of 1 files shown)


skddc:
> In conjunction with #952 it also sends the network-online event.

It already did send the event in that branch, when manually restarting sync. Not sure what this means.

skddc:
> Also, same as I mentioned for the Google Drive client, after coming back online the widget still says "Offline". So far I haven't found the cause for that.

I asked about this in #952. Also don't understand what that means.

skddc:
Works exactly as advertised.

+1

Issue #957, opened 2016-11-26
Dropbox and Google Drive don't send `wire-busy` and `wire-done` events


galfert:
As mentioned [here](https://github.com/remotestorage/remotestorage.js/pull/952#issuecomment-263057975), the events are not sent although they [are announced](https://github.com/remotestorage/remotestorage.js/blob/6a397f393ebb9b27d869640f6bc474d7dc8f5fd1/src/dropbox.js#L196).

PR #958, opened 2016-11-29
Send wire-busy and wire-done events for Dropbox and Google Drive

galfert:
refs #957

Should be merged after #952, as there might be merge conflicts.


galfert:
Replaced by #959.

PR #960, opened 2016-11-30
Make Travis test on several Node versions

galfert:
This adds versions 0.12.x, latest 6.x and latest node (currently 7.x) to be tested on Travis. Was version 0.10.x only before.


galfert:
Added a fix for the failing spec on Node 7.x.

skddc:
Nice one! +1

PR #961, opened 2016-12-03
Fetch Google Drive user name

### LINKED ISSUE #955: Set user address (or name) for Google Drive connections
Currently the userAddress is not set, so it's not possible to know/show what account the current token is for.

Dropbox is setting an actual full name of the account holder from somewhere. We need to check what we can get from Google and how.

galfert:
Closes #955 

Fetches the user info on connect and caches the user address in localStorage.

Also fixes the caching of the user address for Dropbox.

### DIFF
--- a/src/googledrive.js
+++ b/src/googledrive.js
@@ -231,7 +231,7 @@
      *   A promise to the user's info
      */
     info: function () {
-      var url = BASE_URL + '/drive/v2/about';
+      var url = BASE_URL + '/drive/v2/about?fields=user';
       // requesting user info(mainly for userAdress)
       return this._request('GET', url, {}).then(function (resp){
         try {
--- a/src/googledrive.js
+++ b/src/googledrive.js
@@ -125,6 +125,14 @@
         }
       };
 
+      var handleError = function() {
+        this.connected = false;
+        delete this.token;
+        if (hasLocalStorage) {
+          localStorage.removeItem(SETTINGS_KEY);
+        }
+      };
+
       if (this.token) {
         this.connected = true;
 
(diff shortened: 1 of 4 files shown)


src/googledrive.js
```diff
@@ -180,6 +221,28 @@
...
+      var url = BASE_URL + '/drive/v2/about';
+      // requesting user info(mainly for userAdress)
+      return this._request('GET', url, {}).then(function (resp){
+        try {
+          var info = JSON.parse(resp.responseText);
+          return Promise.resolve(info);
+        } catch (e) {
+          return Promise.reject(e);
+        }
+      });
```
skddc: What if this request fails, i.e. the Promise is rejected?

test/unit/dropbox-suite.js
```diff
@@ -290,6 +296,30 @@ define(['bluebird', 'requirejs', 'test/behavior/backend', 'test/helpers/mocks'],
       },
 
       {
+        desc: "#configure fetches the user info when no userAddress is given",
+        run: function (env, test) {
+          env.rs.widget = { view: { setUserAddress: function() {} } };
+
+          env.client.on('connected', function() {
+            test.assert(env.client.userAddress, 'John Doe');
```
skddc: This is not really the user address. Can we get both the name and email address for the account maybe?

galfert:
I changed the user info request to only ask for the actual user info so it's less data that needs to be transferred.

Will try to change error handling next.

As for the `userAddress`, I think it's best to use the email address instead. When you have multiple Google Drive accounts, they probably have the same `displayName`. And combining them to e.g. `"John Doe (john.doe@gmail.com)"` might be a bit long for showing it in the widget.

Opinions?

skddc:
> I changed the user info request to only ask for the actual user info so it's less data that needs to be transferred.

w00t! Perfect.

> As for the userAddress, I think it's best to use the email address instead. When you have multiple Google Drive accounts, they probably have the same displayName. And combining them to e.g. "John Doe (john.doe@gmail.com)" might be a bit long for showing it in the widget.
> 
> Opinions?

Good point. I also think that's the way to go.

galfert:
The `userAddress` is set to the email address from the Google Drive info now.
Also when the info couldn't be fetched, an error event is emitted. Did the same for the Dropbox client as well.

galfert:
Found out that the Dropbox user info also contains the email address. So for the same reasons as for Google Drive and for the sake of consistency, I changed the Dropbox client to also use the email address as `userAddress`.

skddc:
Works as advertised and LGTM.

I'll merge this without a second review, because it's only changing code for experimental features, and I'm using this for the `remove_widget` branch and the new widget repo already.

PR #963, opened 2016-12-17
[WIP] SAFE backend

theWebalyst:
This is just to show where I am - not intended for merging - but so you can see what is involved in a backend for SAFE Network. I have had myfavoritedrinks, litewrite and Laverna demos all working well, though they may not be with this exact code! As soon as I have them reliable I'll update everyone.

Currently this may work with those demos to some extent. Widget not showing states, not sure if sync is working fully right now though it was recently. Definitely don't use with real data!

The SafeNetwork backend requires using SAFE Launcher and SAFE Browser to access demos on SAFE Network and some extra magic to run locally for testing. Ask if interested! 

SAFE Network is in alpha and parts of this changing a lot right now.


src/remotestorage.js
```diff
@@ -270,7 +270,7 @@
   };
 
   RemoteStorage.config = {
-    logging: false,
+    logging: true,//mrhTODO disable logging
```
skddc: This shouldn't be committed and ideally be done from the app/client code instead of here.

src/remotestorage.js
```diff
@@ -538,14 +538,32 @@
...
+     * mrhTODO the following is changing needs updating once new scheme clear
+     * 
+     * Note that SafeNetwork (SAFE Network) backend does not require an API key from
+     * an authority as in the case of most cloud services, so setApiKeys() is
+     * used to define the application related information that is passed to 
+     * an authorisation gateway (SAFE Launcher). The gateway can return a stored
+     * authorisation token or may asks the user for permission to generating one.
+     * 
+     * See also the 'backends' example in the starter-kit (deprecated). Note that 
+     * support for both these backends is still experimental.
```
skddc: This function is used for setting API keys, as the name suggests. It doesn't make sense to use it for anything else. Either add a new function that expresses what is set for SN, or add more arguments to `connect`.

src/remotestorage.js
```diff
@@ -721,7 +744,7 @@
       }
 
       function featureSupported(name, success) {
-        self.log("[RemoteStorage] [FEATURE "+name+"]" + success ? "":" not"+" supported");
+        self.log("[RemoteStorage] [FEATURE "+name+"]" + (success ? "" : " not") + " supported");
```
skddc: This looks like an actual fix for a bug here, right? I think there's a space char too much in front of the colon, but other than that it should probably go in a pull request to master.

src/sync.js
```diff
@@ -1,7 +1,7 @@
 (function (global) {
 
-  var syncInterval = 10000,
-      backgroundSyncInterval = 60000,
+  var syncInterval = 10000, // mrhTODO re-instate: 10000,
+      backgroundSyncInterval = 60000, // mrhTODO re-instate: 60000,
```
skddc: These comments look obsolete.

skddc:
Just FYI: current alpha is in master, and will be published as beta very soon. You could rebase on top of that, if you like.

theWebalyst:
Thanks @skddc I'm sure you've seen the update re-Safenetwork (my favourite drinks up on Alpha 2 which launched yesterday) and yes, I'll want to rebase before long... how long? I'm not sure 😄 and haven't yet looked at what it involves. First though I would like to get Litewrite up, see if Laverna still throws errors for me. Then probably rebase etc.

Its good to see RS coming along and getting ready for beta. I still haven't played with the new widget but looking forward to it.

skddc:
I'm closing this PR, because it's too old. The current version of rs.js is 1.2.1, so this would have to be rebased and opened in a new PR in any case.

That's not to say that the feature is rejected, of course! Just triaging the PR list with this.

theWebalyst:
It is out of date on the SAFE side as well, so correct to close. Also, I understand it was not regarded as desirable within RS, but preferred to remain as a fork. I'm busy with other stuff, but some of that (Safenetworkjs) will make it easy for me to update this should it be desired.

So if anyone sees this and thinks "shame, that would be cool" let me know. Good luck everyone.

PR #974, opened 2017-04-25
Feature/968 loading modules

### LINKED ISSUE #968: New API for loading modules
This is taken from https://community.remotestorage.io/t/ideas-about-api-changes-additions-in-noglobal-1-0/380.

Currently loading the modules happen implicitly by just importing them. This requires an inherent order for the imports, in that the RemoteStorage library has to be imported first.

The proposed change for the module loading could look something like

```js
import Bookmarks from 'remotestorage-module-bookmarks';
import Documents from 'remotestorage-module-documents';
import RemoteStorage from 'remotestoragejs';

let rs = new RemoteStorage({ modules: [Bookmarks, Documents] });
```

This way it would also be possible to use several remoteStorage instances that use different modules.

lesion:
this should fix #968


skddc:
I'm looking into this next, but just one comment from my first glance: we shouldn't change version numbers or tag releases in a branch or PR. This has to be done in `master`, after sth has been merged.

skddc:
I was wondering, for this piece of code to work, we'll have to change the exports of the modules, right?

```js
import Bookmarks from 'remotestorage-module-bookmarks';
import Documents from 'remotestorage-module-documents';
import RemoteStorage from 'remotestoragejs';

let rs = new RemoteStorage({ modules: [Bookmarks, Documents] });
```

The way I understand it, they would have to export an object now, instead of just the builder function? Like so e.g.:

```js
module.exports = {
  name: "bookmarks",
  builder: function(privClient, pubClient){}
}

Is that correct?

skddc:
I can confirm that this works with the new API (aside from running into an issue with my imported object containing an extra `default` key that my module loader didn't recognize automatically). But it definitely did not work with the old API for the same module. The property wasn't available on the `remoteStorage` instance then.

@lesion Could you confirm that this still works with the old API for you?

As already discussed briefly on IRC, I'd also be fine with axing support for the old API entirely, so that we don't lug around some extra legacy there. @galfert What do you think?

galfert:
Might actually be beneficial to have one more reason to port the modules :)

skddc:
Ok, @lesion if you agree as well (iirc you indicated it on IRC already), then want to remove the backwards compat?

michielbdejong:
Sure, go ahead! :+1:

silverbucket:
Yeah I think we're still in a good position to break API and shed dead weight. Hopefully not the case going forward :)
👍

skddc:
Thanks everyone! I went ahead and removed the legacy API.

As we decided to suspend the two-plus-one rule on `master` until it is the actual stable branch again, I'll rebase this locally to clean up the history a bit and merge then.

Good job, @lesion!

group-five

PR #975, opened 2017-04-27
Migrate Dropbox back-end to Dropbox API version 2

### LINKED ISSUE #936: Migrate to DropBox API v2
API documentation: https://www.dropbox.com/developers/documentation/http/documentation
Migration guide: https://www.dropbox.com/developers/reference/migration-guide

It seems that all the required functionality is present in the API v2.

Ragnis:
connected to #936
closes #936

### DIFF
--- a/src/dropbox.js
+++ b/src/dropbox.js
@@ -322,35 +322,63 @@
      */
     _getFolder: function (path, options) {
       // FIXME simplify promise handling
-      var url = 'https://api.dropbox.com/1/metadata/auto' + getDropboxPath(path);
+      var url = 'https://api.dropboxapi.com/2/files/list_folder';
       var revCache = this._revCache;
       var self = this;
 
-      return this._request('GET', url, {}).then(function (resp) {
-        var status = resp.status;
-        if (status === 304) {
-          return Promise.resolve({statusCode: status});
+      var processResponse = function (resp) {
+        var body, listing;
+
+        if (resp.status !== 200) {
+          return Promise.reject('Unexpected response status: ' + resp.status);
         }
-        var listing, body, mime, rev;
-        try{
+
+        try {
           body = JSON.parse(resp.responseText);
         } catch (e) {
           return Promise.reject(e);
         }
-        rev = self._revCache.get(path);
-        mime = 'application/json; charset=UTF-8';
-        if (body.contents) {
-          listing = body.contents.reduce(function (m, item) {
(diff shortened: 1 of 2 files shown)


src/dropbox.js
```diff
@@ -381,31 +411,49 @@
...
         var meta, body, mime, rev;
-        if (status !== 200) {
+        if (status !== 200 && status !== 409) {
           return Promise.resolve({statusCode: status});
         }
-
+        meta = resp.getResponseHeader('Dropbox-API-Result');
         body = resp.responseText;
+
+        if (status == 409) {
```
skddc: Is JSHint missing from the Travis runs now? I think that would complain on my machine.

src/dropbox.js
```diff
@@ -543,11 +585,11 @@
             });
           }
 
-          return self._deleteSimple(path);
+          return this._deleteSimple(path);
```
Ragnis: I've now replaced the parent function with an arrow function.

galfert:
I think we are almost there :)

One last thing I found when testing it with Webmarks:

When trying to delete the last bookmark, it gets deleted locally, but I can see the item still in my Dropbox. When disconnecting and reconnecting I also see the Bookmark in Webmarks again. But then when trying to delete it again, the `remove` call's promise fails with an error

> Cannot delete non-existing node / 

@Ragnis please let me know if you need any more info to be able to debug this.

skddc:
Maybe there's some kind of trashbin status and it's possible to retrieve files that have just been deleted? Just a wild guess.

galfert:
> Maybe there's some kind of trashbin status and it's possible to retrieve files that have just been deleted? Just a wild guess.

Yes, I think you can restore files in Dropbox in the Web interface. But all the other files got deleted normally, so I don't think that's the reason.

skddc:
@Ragnis Did you see the latest comments in here? Let @galfert know if/how he could help.

skddc:
@galfert It looks to me like the review comment, which prevents merging this, has been addressed, but that you also found a new issue?

galfert:
Ok, I think the remaining error I'm seeing when trying to delete the very last file in storage isn't actually related to or only happening in this PR. I also get the same error with the Google Drive backend.

So I think it's best to merge this and create a separate issue for the other error.

@Ragnis thanks again for this great upgrade. And sorry for leaving this unmerged for so long.

galfert:
Replaced by #992.

PR #984, opened 2017-05-29
Add "name" property to custom error objects

skddc:
This adds the property `name` with the error name for our custom errors, so that when handling them one can check for `error.name` instead of `error instanceof Discovery.DiscoveryError` for example. The latter is not possible when importing `RemoteStorage` from the module/package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "version": "1.0.0-alpha1",
   "private": false,
   "license": "MIT",
-  "main": "./release/head/remotestorage.js",
+  "main": "./release/stable/remotestorage.js",
   "repository": {
     "type": "git",
     "url": "git://github.com/remotestorage/remotestorage.js.git"
--- a/src/authorize.js
+++ b/src/authorize.js
@@ -1,6 +1,6 @@
 var log = require('./log');
 var util = require('./util');
-  
+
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -4,7 +4,7 @@
   "version": "1.0.0-alpha1",
   "private": false,
   "license": "MIT",
-  "main": "./release/head/remotestorage.js",
+  "main": "./release/stable/remotestorage.js",
```
galfert: Shouldn't we wait with this change until we do an actual release?

galfert: Nevermind, just saw the commit description.

galfert:
👍  lgtm and tested.

skddc:
Thanks!

I'm merging this with only one review now. I think we can/should suspend the two-plus-ones rule on `master`, while we're actually using it as a development branch, same as we also don't have it for any other branches than the stable one. And the stable branch is currently "stable", so let's try to move towards the 1.0 beta as quickly as possible (while still doing peer reviews of all changes). Sound good?

galfert:
> So let's try to move towards the 1.0 beta as quickly as possible (while still doing peer reviews of all changes). Sound good?

Sounds all good to me 👍

PR #987, opened 2017-08-04
Fix Google Drive info trying to update the widget

galfert:
Closes remotestorage/remotestorage-widget#13

There is no access to `rs.widget` anymore. The new widget uses the `userAddress` property proactively now.

### DIFF
--- a/src/googledrive.js
+++ b/src/googledrive.js
@@ -145,7 +145,6 @@
         } else {
           this.info().then(function(info) {
             this.userAddress = info.user.emailAddress;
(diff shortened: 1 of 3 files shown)


skddc:
This works, but it's showing my email address instead of my name. Is that expected?

galfert:
Yes, we decided to do it the same way as with Dropbox. You might have different accounts using the same name, but the email address will be different.

skddc:
Cool. We'll just have to change the icon to be the one for the connected backend, but that's a different issue.

PR #988, opened 2017-08-07
Allow to configure network request timeout

### LINKED ISSUE #983: Add config for WireClient.REQUEST_TIMEOUT
Used to be possible to configure by just changing the variable on the class, but with the new layout, we need an API for it. That's also the much cleaner way of doing it anyway.

galfert:
Closes #983

Can be set via `setRequestTimeout` and retrieved via `getRequestTimeout`.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -13,6 +13,7 @@ var config = {
   backgroundSyncInterval: 60000,
   isBackground: false,
   cordovaRedirectUri: undefined,
+  requestTimeout: 30000,
   modules: []
 };
 
--- a/src/remotestorage.js
+++ b/src/remotestorage.js
@@ -171,7 +171,7 @@
   };
 
 
-  // TOFIX: Instead of doing this, would be better to only 
+  // TOFIX: Instead of doing this, would be better to only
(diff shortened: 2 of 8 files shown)


skddc:
Nice!

Did you try this in Sharesome already?

galfert:
> Did you try this in Sharesome already?

No, so far only in the widget demo.

skddc:
The widget demo doesn't upload or download large files, so that wouldn't really make sure the timeout works, no?

galfert:
Since the actual usage of the timeout didn't change, I just added a `console.log` right before where the setting is used to make sure it's using the correct value.

But I will also try it out with Sharesome.

galfert:
I tried it with Sharesome now. First without configuring it explicitly and uploading a large file I got a timeout. Then I configured the timeout to be 90000ms and the upload worked.

I will create a PR for Sharesome once this is published in a release.

skddc:
Awesome!

LGTM

skddc:
@galfert Could you update the CHANGELOG accordingly?

galfert:
Updated the CHANGELOG in 94dda87.

skddc:
Cool, thx.

PR #989, opened 2017-08-08
Fix discover tests

galfert:
I noticed that the Travis tests were failing for a while now. It was two of the test cases of the `Discover` suite that failed because of a timeout. 

I traced this down to a difference in the version of `webfinger.js`. Travis was using version 2.6.5, while locally I still had version 2.6.4. And the tests were running fine locally.

So far, I don't quite know if it's the way we mock the network requests wrong in the tests or if it's actually a regression in `webfinger.js`.

Maybe @silverbucket can chime in about what exactly changed between the two versions?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3332,14 +3332,16 @@
       }
     },
     "webfinger.js": {
-      "version": "https://registry.npmjs.org/webfinger.js/-/webfinger.js-2.6.4.tgz",
+      "version": "2.6.4",
+      "resolved": "https://registry.npmjs.org/webfinger.js/-/webfinger.js-2.6.4.tgz",
       "integrity": "sha1-rDJs9ej/Qo+BHSsd0Y6VrlFvTf4=",
       "requires": {
-        "xhr2": "https://registry.npmjs.org/xhr2/-/xhr2-0.1.4.tgz"
+        "xhr2": "0.1.4"
       },
       "dependencies": {
         "xhr2": {
-          "version": "https://registry.npmjs.org/xhr2/-/xhr2-0.1.4.tgz",
+          "version": "0.1.4",
+          "resolved": "https://registry.npmjs.org/xhr2/-/xhr2-0.1.4.tgz",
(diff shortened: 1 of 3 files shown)


silverbucket:
@galfert Hmm, this is odd, because I have the exact same test (calling `foo@bar` and expecting a failure callback) in the webfinger.js test suite specifically to match the remotestorage one. in the webfinger one I get a failure callback, in the RS one it times out...

galfert:
I think it might have to do with the way the `XMLHTTPRequest` calls are mocked. I don't understand it yet, but the mocks seem to not only be relevant for the one specific test case they are defined in, but it looks like they leak into the other test cases as well.

silverbucket:
@galfert there was a bugfix submitted to webfinger.js which is what prompted the 2.6.5 release, specifically dealing with behavior where a callback was called twice in some cases. It seems that this bug was actually covering up for whatever issue is going on with the XMLHttpRequest mock.

silverbucket:
Figured out the issue is when a url fails, and fallback is enabled, there are a series of different paths that are tried, each of these is a new request, so the tests should access the latest `xhr` instance to initialize the `onreadystatechange()` otherwise the same (old) handlers are called, which now are ignored, and we get a timeout. Pushing fixed test now.

silverbucket:
@galfert if it makes sense to you go ahead and merge it. I just wanted to leave it open until someone had a look.

galfert:
Cool, thanks. I will check it out.

@silverbucket One thing I was wondering about when I looked at that fix in `webfinger.js`, is that it contains a return without any value. But all other ways the `__processState` method exits, is by either returning a `successHandler` or a `errorHandler`. Do you think that's correct?

silverbucket:
@galfert yeah this is in the case that the same handler is called a second time, it should never call the same callback twice. Unless I'm missing something?

galfert:
Works great, thanks again.

I merged it manually to remove my superfluous commit that downgraded `webfinger.js`.

PR #992, opened 2017-08-23
Dropbox v2 (rebased)

galfert:
This is the continuation of PR #975

It's rebased on the latest master and I squashed some of the minor (typo) fix commits.


skddc:
Note: as an app developer you have to explicitly allow API v2 for existing apps.

skddc:
Works like a charm!

@Ragnis 👏

Issue #972, opened 2017-04-17
Document the new library usage (import/require/global/... + new widget)


skddc:
I think this should finally be done in the generated API docs, which are currently missing the whole intro for how to use the library.

galfert:
Also document that users need to bring their own polyfill for promises now if they want to support older browsers.

skddc:
I learned how Read the Docs works, and created a basic Sphinx setup with the RtD theme for our `doc` folder. I also converted the first two files to `.rst` in there, and thought about how to structure the docs. There's a special `Contributing` section now, so we can keep information for contributors seperate from library usage docs for app developers for example (and thus also create way more docs for contributing).

skddc:
I signed up for Read the Docs and imported the branch, so it should be built automatically when pushing to that now: https://remotestoragejs.readthedocs.io/en/docs-sphinx/

The "edit on GitHub" link is wrong, because it has issues with slashes in branch names (they're converted to dashes over there).

skddc:
Last update for today: I converted all existing files in `doc` to RST now (pandoc FTW!) and added all of them to the menu: https://remotestoragejs.readthedocs.io/en/docs-sphinx/

Automatic builds on push work like a charm, so anyone can now edit and create docs in that branch.

I think the next step would be adding a new "getting started" section and afterwards seeing how much of the existing content is still useful when editing it and what has to be deleted or rewritten entirely.

Issue #801, opened 2014-10-15
Investigate switching to a more suitable documentation framework


skddc:
NaturalDocs' last release is older than this whole library, and we're running into limitations regularly. The problem with not being able to declare methods as private just popped up again on IRC, and as we've been talking about this multiple times now, it looks like we should finally collect options and their pros and cons, so we can progress towards a decision at some point.

skddc:
Someone just pointed me to [documentation.js](http://documentation.js.org/), which is based on [JSDoc](https://github.com/jsdoc3/jsdoc).

skddc:
We're switching to JSDoc plus Sphinx.

Issue #785, opened 2014-10-10
Improve documentation of change events


skddc:
The change event documentation is extremely confusing at the moment: http://remotestorage.io/doc/code/files/baseclient-js.html#RemoteStorage.BaseClient.change

skddc:
I actually didn't change much about the event API docs and examples when moving them to the RST page. Closing this for now, not withstanding the need for more editing/improvements in case people still find them confusing when reviewing the new docs.

Issue #748, opened 2014-08-20
Things that are confusing to newcomers


skddc:
From IRC (and exactly what I said about it before):

> I can't understand why there is a "remoteStorage" and a "RemoteStorage", neither when should I use these "ready", "connected", "disconnected" callbacks

The latter just needs better documentation I believe, whereas the former needs a better lib architecture at some point in the future I think (maybe we can hack it for now, though?).

Just put this here, so it doesn't disappear and we can decide what to do with the feedback.

skddc:
Closing in favor of #972

PR #997, opened 2017-10-02
Switch documentation to JSDoc and Sphinx

### LINKED ISSUE #864: Sync interval getter/setter functions hidden in internal API docs
Due to the sync interval docs being contained in the sync file, but creating functions on the `RemoteStorage` prototype, they don't appear on the `RemoteStorage` page as public API functions.

skddc:
I've been working on considerably improving the documentation situation in recent weeks. It's now at a point, where I think other people can and should have a look at it, and any feedback or contributions would be most appreciated.

You can find all relevant information about how the new docs are written and generated in the new docs themselves: https://github.com/remotestorage/remotestorage.js/blob/docs/sphinx/doc/contributing/docs.rst

### Warning

Unfortunately, one part of the puzzle is still somewhat missing: Read the Docs haven't updated their build images on all production workers yet, so the building and publishing to what is supposed to be the new public documentation website at https://remotestoragejs.readthedocs.io/ fails most of the time. (See https://github.com/rtfd/readthedocs.org/issues/3069 for details.)

refs #972, #801, #785, #748, #732, #707
closes #864


galfert:
I don't seem to have the `sphinx-autobuild` command that is mentioned in `doc/contributing/docs.rst`.

galfert:
`sphinx-autobuild` seems to be a separate requirement. I'll add it to `doc/requirements.txt`.

galfert:
I also had to add `pip install sphinx_rtd_theme` to the `requirements.txt`, otherwise I get this error when running `sphinx-autobuild`:

```
Theme error:
sphinx_rtd_theme is no longer a hard dependency since version 1.4.0. Please install it manually.(pip install sphinx_rtd_theme)
```

skddc:
@galfert Great, thanks!

I also just moved all untouched legacy docs to the same folder/namespace, so that they're all in one place.

galfert:
Building and serving the docs works for me now. I just get this warning:

```
WARNING: html_static_path entry u'remotestorage.js/doc/_static' does not exist
```

And when browsing the documentation, I can see this in the logs:

```
[W 171004 16:22:56 web:2063] 404 GET /_static/fonts/fontawesome-webfont.woff2?v=4.6.3 (127.0.0.1) 4.01ms
```

Other than that it looks good.

skddc:
You can just create that `_static` directory in `doc/` to eliminate the warning (will be ignored). Looks like an issue with the RTD theme extension.

skddc:
@galfert I think it would be a good idea if you ported the rest of the public JS API (`caching`, `access`, not sure what else) to the new docs, in order to get accustomed to the new setup. Then we have at least two people who can help other contributors with it from here on.

skddc:
By the way, we've been lucky with the RTD builds today. Some still failed, but the last 3 passed, so whoever's interested can also preview the current state of the new docs over there:

http://remotestoragejs.readthedocs.io/

galfert:
> @galfert I think it would be a good idea if you ported the rest of the public JS API (caching, access, not sure what else) to the new docs, in order to get accustomed to the new setup. Then we have at least two people who can help other contributors with it from here on.

Ok, I will give it a try.

src/remotestorage.js
```diff
@@ -1,829 +1,792 @@
...
+  onChange: function (path, handler) {
+    if (! this._pathHandlers.change[path]) {
+      this._pathHandlers.change[path] = [];
     }
+    this._pathHandlers.change[path].push(handler);
+  },
 
-  };
+  /**
+   * TODO: do we still need this, now that we always instantiate the prototype?
```
galfert: Not sure I get the question, but I think we still need this for enabling/disabling the logging during runtime (i.e. after initial creation of the remoteStorage instance).

skddc: Yes, but why would you do that? Or, in other words, did you ever do that yourself? And if so, did you not just do it, because you couldn't instantiate the instance?

galfert: I can't really anticipate all use-cases others might have. I myself used it a few times for debugging, enabling the log for a certain action and then disabling it again so it's not so noisy.

skddc: It's not really about anticipating, but more like giving away clothes you haven't used in the last 12 months. Enabling the logs for a certain action and then disabling before reload seems like a rather rare use case to me.

However, adding this function instead of the init config, and then losing lots of logs without realizing it (which happened to me every time I used it myself) seems like an actual problem we should try to not have people run into.

skddc:
@galfert I'm pretty sure everything but `claim()` is private for `Access`. If we actually want people to manually manage the prototype's access config, we might at least want to add notes, or a second header, in order to ensure nobody thinks that these are all necessary or reasonable to integrate.

doc/js-api/caching.rst
```diff
@@ -0,0 +1,46 @@
...
+-------------------
+
+.. autofunction:: Caching#set
+  :short-name:
+
+  Example:
+
+  .. code:: javascript
+
+     remoteStorage.caching.set('/bookmarks/archive')
```
skddc: I think this is missing a strategy, because usually you'd use `enable()` if you don't have any explicit setting in mind.

doc/js-api/caching.rst
```diff
@@ -0,0 +1,46 @@
...
+
+.. autofunction:: Caching#enable
+  :short-name:
```
skddc: Example would be nice.

skddc:
@galfert Are we actually certain that the caching strategy options are up to date?

doc/js-api/caching.rst
```diff
@@ -0,0 +1,46 @@
+Caching
+=======
```
skddc: I think an introductory note that is initialized as `remoteStorage.caching`, but only if one doesn't configure `caching: false`, would be useful.

galfert:
> @galfert Are we actually certain that the caching strategy options are up to date?

I couldn't find anything that would indicate otherwise.

skddc:
Note: we decided in the IRC meeting yesterday to merge this asap. I don't see why this couldn't live in master tbh, so one can branch off from that again for all the remaining docs issues. Removing WIP. Still needs rebase tho.

PR #998, opened 2017-10-04
Stop all sync requests after disconnect

### LINKED ISSUE #939: Sync error after disconnect
There seems to be some race condition where a sync task is still scheduled and executed after disconnecting storage, but the DB is not available anymore:

screenshot from 2016-06-21 18-35-56

It doesn't really break this app (and shouldn't break other apps). Only throws an exception after logout in some cases. We should delete all sync tasks during disconnect anyway.

### LINKED ISSUE #994: Additional sync request after disconnecting from Dropbox
I haven't noticed that before, but when disconnecting from Dropbox, remoteStorage still tries one last sync request afterwards that fails with a 400 (Bad Request) error. The widget then shows this:

screen shot 2017-08-25 at 12 35 12

I could not produce the same problem with neither remoteStorage nor the Google Drive backends.

galfert:
Fixes #994

This stops all scheduled sync requests after disconnecting.

During testing I noticed that for remoteStorage and Google Drive backends it would also still do one `sync()` call after disconnecting, but that one would not do a request for a directory listing like for the Dropbox backend. So this change might indeed also fix #939 as @skddc suspected, but I'm not sure about this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,10 @@ All releases can also be found and downloaded on the
 
 * Switch Dropbox adapter to use the Dropbox API v2 (#936)
 
+### Bugfixes
+
+* Remove all scheduled sync calls after disconnect (#994)
+
 ## 0.14.0 (November 2016)
 
 ### Enhancements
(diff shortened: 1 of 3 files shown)


skddc:
+1 LGTM!

I'm pretty much certain that this fixes #939 as well.

skddc:
@galfert Could you add a CHANGELOG entry for this?

galfert:
I added the Changelog entry.

I'd propose to close #939 as well when merging this with a comment to re-open if it still happens.

PR #1001, opened 2017-10-05
Remove deprecations

galfert:
I think a major release is probably the right time to remove old deprecations and not just warn about them anymore.

The merge target is the `docs/sphinx` branch, just to prevent merge conflicts later on.


galfert:
I also removed a deprecated method `equalObj` from `util`.

Issue #854, opened 2015-04-02
GD API docs


skddc:
There's nothing at all in https://remotestorage.io/doc/code/files/googledrive-js.html

skddc:
These have actually been added by @galfert in the linked PR.

PR #1009, opened 2017-10-06
Store Google Drive data under "remotestorage" dir

### LINKED ISSUE #962: Store GDrive data in "remoteStorage" base dir
Currently, category dirs are created on the base of a Google Drive account. I think we need to store all data in a common RS base dir, same as we recently introduced it for Dropbox. Otherwise we risk clashes and bugs as soon as people really start using it.

galfert:
Closes #962
refs #854

The feature itself works already. The PR is still WIP, because I'm still cleaning/refactoring the code and adding documentation.

### DIFF
--- a/src/googledrive.js
+++ b/src/googledrive.js
@@ -28,7 +28,7 @@ var BASE_URL = 'https://www.googleapis.com';
 var AUTH_URL = 'https://accounts.google.com/o/oauth2/auth';
 var AUTH_SCOPE = 'https://www.googleapis.com/auth/drive';
 var SETTINGS_KEY = 'remotestorage:googledrive';
-var PATH_PREFIX = '/remoteStorage';
+var PATH_PREFIX = '/remotestorage';
 
 var GD_DIR_MIME_TYPE = 'application/vnd.google-apps.folder';
 var RS_DIR_MIME_TYPE = 'application/json; charset=UTF-8';
--- a/src/googledrive.js
+++ b/src/googledrive.js
@@ -19,23 +19,24 @@
  * Docs: https://developers.google.com/drive/web/auth/web-client#create_a_client_id_and_client_secret
  **/
 
-var Authorize = require('./authorize');
-var WireClient = require('./wireclient');
-var eventHandling = require('./eventhandling');
-var util = require('./util');
+const Authorize = require('./authorize');
+const WireClient = require('./wireclient');
+const eventHandling = require('./eventhandling');
+const util = require('./util');
 
-var BASE_URL = 'https://www.googleapis.com';
-var AUTH_URL = 'https://accounts.google.com/o/oauth2/auth';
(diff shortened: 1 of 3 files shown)


skddc:
Great!

Just one comment so far: In the Dropbox backend, the prefix is the lowercase version `remotestorage`, which I think is better for something synced across filesytems.

skddc:
I added a ref to #854 and assigned that one as well.

galfert:
> Just one comment so far: In the Dropbox backend, the prefix is the lowercase version remotestorage, which I think is better for something synced across filesytems.

I'll change it to all lowercase as well.

src/googledrive.js
```diff
@@ -114,13 +142,22 @@ GoogleDrive.prototype = {
...
 
+  /**
+   * Configure the Google Drive backend.
+   *
+   * Fetches the user info from Google when no ``userAddress`` is given.
+   *
+   * @param {object} settings
+   * @param {string} [settings.userAddress] - The user's email address
+   * @param {string} [settings.token] - Authorization token
+   */
```
skddc: I think this is a private method, because a developer should only use `connect()` and have the library deal with the rest, no?

src/googledrive.js
```diff
@@ -144,31 +181,46 @@ GoogleDrive.prototype = {
...
       this._emit('not-connected');
     }
   },
 
+  /**
+   * Request a resource (file or directory).
+   *
+   * @param {string} path - Path of the resource
+   * @param {object} options - Request options
+   * @returns {Promise} Resolves with an object containing the status code, body, content-type and revision
```
skddc: This is also private I think. A developer should only use `BaseClient` methods for get/put/delete calls.

src/googledrive.js
```diff
@@ -177,20 +229,31 @@ GoogleDrive.prototype = {
...
 
+  /**
+   * Create or update a file.
+   *
+   * @param {string} path - File path
+   * @param body - File content
+   * @param {string} contentType - File content-type
+   * @param {object} options
+   * @param {string} options.ifNoneMatch - Only create of update the file if the current ETag doesn't match this string
+   * @returns {Promise} Resolves with an object containing the status code, content-type and revision
```
skddc: See `get`.

src/googledrive.js
```diff
@@ -177,20 +229,31 @@ GoogleDrive.prototype = {
...
   },
 
+  /**
+   * Create or update a file.
+   *
+   * @param {string} path - File path
+   * @param body - File content
+   * @param {string} contentType - File content-type
+   * @param {object} options
+   * @param {string} options.ifNoneMatch - Only create of update the file if the current ETag doesn't match this string
```
skddc: It's perfectly possible to break lines in these descriptions. See other source files.

galfert:
The commented issues should be fixed.

I removed the [WIP], I think this is ready for merge now.

skddc:
👍

PR #1011, opened 2017-10-09
[WIP] Documentation and refactorings

galfert:


src/dropbox.js
```diff
@@ -354,14 +354,14 @@
...
      * Checks for the path in ``_revCache`` and decides based on that if file
-     * has changed. Calls ``_getFolder`` is the path points to a folder.
+     * has changed. Calls ``_getFolder`` if the path points to a folder.
      *
      * Calls ``Dropbox.share`` afterwards to fill ``_itemRefs``.
      *
      * Compatible with ``WireClient.get``
      *
-     * Checks for the path in _revCache and decides based on that if file has
-     * changed. Calls _getFolder if the path points to a folder.
```
skddc: Why delete the explanation of what the function does?

gregkare: It's not deleted, it was already above. The Git merge looks weird because I fixed a typo as well (s/is/if)

src/remotestorage.js
```diff
@@ -703,16 +703,18 @@ RemoteStorage.prototype = {
...
     if (!config.cache) { return; }
     this.sync.stopped = false;
     this.syncStopped = false;
-    this.sync.sync();
+    return this.sync.sync();
   },
 
   /**
-   * TODO: document
+   * Stop the synchronization interval.
```
skddc: One can't "stop" an interval really. Is this referring to stopping the periodic sync altogether?

galfert: Yes.

src/remotestorage.js
```diff
@@ -707,14 +707,17 @@ RemoteStorage.prototype = {
    * @returns {Promise} A Promise which resolves when the sync has finished
    */
   startSync: function () {
-    if (!config.cache) { return; }
+    if (!config.cache) {
+      console.warn('Can not start syncing, because caching is disabled.');
```
skddc: I'd say "Nothing to sync" instead, but no strong opinion. If we leave it like this, grammatically more appropriate would be "cannot" without a space.

PR #1012, opened 2017-10-09
Remove change event from remote clients

### LINKED ISSUE #1007: Fix or remove change event on `WireClient`
There's a code comment about change events on the wireclient not being emitted. We should verify that's the case, and if yes, either remove them (because they're kind of redundant at first glance) or fix them.

galfert:
Fixes #1007

The remote clients are not emitting the change event, that's handled by the baseclient instead.

### DIFF
--- a/src/dropbox.js
+++ b/src/dropbox.js
@@ -191,7 +191,7 @@
       }
     };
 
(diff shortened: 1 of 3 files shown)


skddc:
lgtm

PR #1013, opened 2017-10-11
Remove step that is not required anymore

gregkare:
Travis is using the Makefile now (as of e68d3c45), so no need to update
the `.travis.yml` file when adding new test files


skddc:
lgtm 👍

Issue #1000, opened 2017-10-05
Adjust README.md for new docs


skddc:
Most of it can be linked to the new docs, with just the essential description, pointers, credits, etc. having to stay.

skddc:
https://github.com/LappleApple/feedmereadmes/blob/master/README-maturity-model.md#level-four-readme-with-purpose

PR #1014, opened 2017-10-11
 Update the Makefile and README to use sphinx instead of Natural Docs 

gregkare:
Also remove obsolete Natural Docs files


Makefile
```diff
@@ -42,8 +37,7 @@ remotestorage.js: $(SOURCES)
 	$(NPM) run build
 
 doc:
-	mkdir -p $(DOC_DIR) $(DOC_CONFIG_DIR)
-	$(DOC_BIN) $(DOC_INPUTS) -o html $(DOC_DIR) -p $(DOC_CONFIG_DIR) -s Default $(DOC_CUSTOM_CSS)
+	cd doc && sphinx-build . _build/html
```
skddc: Wouldn't we have to also `cd` back out here? (Haven't tested it, and am not a Make expert.) If so, I think we should rather use relative paths to the root dir for the arguments.

README.md
```diff
@@ -71,8 +71,9 @@ Use the `jaribu` executable in order to test single files, like so e.g.:
 
 ### How to build
```
skddc: This is now part of the docs: https://remotestoragejs.readthedocs.io/en/latest/contributing/docs.html#how-to-build-the-docs-on-your-machine

We also have this issue open for adjusting the README: #1000

Also, wow, I didn't notice when I created issue number 1000. :)

Makefile
gregkare: Good point, I checked and Make invokes a new sub-shell for each line (https://www.gnu.org/software/make/manual/html_node/Execution.html#Execution). At first I tried to use relative paths but I couldn't make it work:

```
$ sphinx-build doc doc/_build/html
Running Sphinx v1.6.4
loading pickled environment... done

Exception occurred:
  File "/usr/local/Cellar/python/2.7.14/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/decoder.py", line 382, in raw_decode
    raise ValueError("No JSON object could be decoded")
ValueError: No JSON object could be decoded
The full traceback has been saved in /var/folders/85/4fncgmr946v2mt7p58hkz6tc0000gn/T/sphinx-err-whlqM5.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
```

skddc: Are you saying we don't have to return then (sounds like it to me)? It'd be easier if you confirmed that explicitly, so we don't have to read the same docs as well.

gregkare: We don't have to return

gregkare:
I'm going to take care of #1000 in this PR since I already started with the README

skddc:
lgtm

PR #1015, opened 2017-10-11
Various versioning/release improvements

skddc:
With proper versioning, package management and CI, we don't need head builds
anymore. Every release meant to be used by developers will be tagged and
published as such.

This adds a bit of automation to the `npm version` command (test first, and if
green then build library and add to the version commit). It also removes the
distinction between `head` and `stable` and only builds a single release to
`release/remotestorage.js`.


skddc:
Ready for review.

gregkare:
It works (macOS Sierra here)! I successfully tried this (`npm version minor`) after doing a temporary local commit to remove the `postversion` to avoid pushing

LGTM

skddc:
Yay! Thanks.

Issue #1003, opened 2017-10-05
Getting Started guide


skddc:
In addition to the JS API docs, we also need a good guide for getting started with rs.js app integration.

* [ ] Adding rs.js to your app
    * [ ] ES module
    * [ ] As global variable
    * [ ] AMD
    * [ ] node.js (rewrite node.js guide)
    * [ ] Note about Promise polyfill (ain't none in rs.js by default anymore)
    * [ ] ...

gregkare:
Now that the Node.js page has been merged this is only missing extra content that we haven't defined yet. What's missing?

skddc:
Actually creating and managing data.

PR #1017, opened 2017-10-12
Document adding rs.js to a project

skddc:
This is still missing an example for AMD, because I can't AMD.

refs #1003


galfert:
LGTM

PR #1018, opened 2017-10-13
Improve/fix authorization error handling

galfert:
This is still WIP, because I'm still working on the Widget-side of things and I don't know yet if there is anything else missing. Also, there are no tests yet.

### DIFF
--- a/src/authorize.js
+++ b/src/authorize.js
@@ -166,7 +166,11 @@ var util = require('./util');
       var authParamsUsed = false;
       if (params) {
         if (params.error) {
-          throw "Authorization server errored: " + params.error;
+          if (params.error === 'access_denied') {
+            throw new Authorize.Unauthorized('User denied access.');
+          } else {
+            throw "Authorization server errored: " + params.error;
+          }
         }
 
         // rsDiscovery came with the redirect, because it couldn't be
--- a/src/authorize.js
+++ b/src/authorize.js
@@ -167,9 +167,9 @@ var util = require('./util');
       if (params) {
         if (params.error) {
           if (params.error === 'access_denied') {
-            throw new Authorize.Unauthorized('User denied access.');
+            throw new Authorize.Unauthorized('Authorization failed: access denied');
           } else {
-            throw "Authorization server errored: " + params.error;
+            throw new Authorize.Unauthorized(`Authorization failed: ${params.error}`);
           }
         }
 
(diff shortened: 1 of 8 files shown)


src/authorize.js
```diff
@@ -166,7 +166,11 @@ var util = require('./util');
       var authParamsUsed = false;
       if (params) {
         if (params.error) {
-          throw "Authorization server errored: " + params.error;
+          if (params.error === 'access_denied') {
+            throw new Authorize.Unauthorized('User denied access.');
```
skddc: How about "Authorization failed: access denied".

src/authorize.js
```diff
@@ -166,7 +166,11 @@ var util = require('./util');
...
+          } else {
+            throw "Authorization server errored: " + params.error;
```
skddc: I think "Authorization failed: [error]" is a bit more concise and human-friendly.

galfert:
I added a fix to not remove any bearer token and/or configuration settings when getting an authorization error. Otherwise, the error wouldn't be shown to users anymore after reloading the app.

skddc:
So is it still WIP?

galfert:
It should all work now, I just didn't look into writing any tests for the changes yet.

skddc:
Ok, cool.

galfert:
I added an optional `code` property to the `Unauthorized` error object.

When a user denies the authorization, the code `access_denied` is added, so the widget can easier detect that particular error.

I tested this in Chrome, Firefox, Safari and Opera. They all can see the additional property on the error. According to [this article](https://medium.com/@xjamundx/custom-javascript-errors-in-es6-aa891b173f87) it shouldn't be a problem to add custom properties to an error object, it just can't have custom methods.

skddc:
That article is about node.js. The potential problems are with e.g. Internet Explorer, Opera Mobile, Android browser, and such.

galfert:
Update: It also works on Edge. But on IE 11 I already get an error during discovery. Still trying to debug that. Looks like it's a problem in Webfinger.js.

galfert:
I fixed the Webfinger.js issue in IE 11 with silverbucket/webfinger.js#27.

After that all worked fine, including the additional error property. I just had to bring my own Promises :)

skddc:
Yay! Good job. 🎉

galfert:
Tested Opera Mobile and Android Browser on Android 4.1.2 and no problems there either.

skddc:
Great! I think we can all agree that's enough then, right?

galfert:
I think so, yes.

galfert:
I added some tests as well, so I think this is ready to merge now.

skddc:
Fantastic!

Issue #1006, opened 2017-10-05
Port remaining docs to JSDoc format


skddc:
I think we're done with the public API, but there are still NaturalDoc comments left for private API, and we should also tag any methods and properties not appearing in the new HTML docs as `@private` or `@protected`.

galfert:
See [my comment](https://github.com/remotestorage/remotestorage.js/pull/997#commitcomment-24783253) about the use of `@private`.

I think we need to distinguish between internal documentaton for library implementors and public documentation for end-users here.

IMO the `@private` label should indicate that the method/property is private to that particular class only, so library implementors know not to use them from anywhere else.

For the user documentation, I think it's good enough to only list those methods that are for public consumption.

skddc:
See my response. There's `@protected` for that.

galfert:
ok

skddc:
🎉

PR #1020, opened 2017-10-13
Port remaining docs to JSDoc format

gregkare:
Connected to #1006

### DIFF
--- a/src/eventhandling.js
+++ b/src/eventhandling.js
@@ -1,12 +1,10 @@
   var log = require('./log');
 
   /**
-   * Interface: eventhandling
+   * @interfacek
    */
   var methods = {
     /**
-     * Method: addEventListener
-     *
      * Install an event handler for the given event name
      */
     addEventListener: function (eventName, handler) {
@@ -22,8 +20,6 @@
     },
 
     /**
-     * Method: removeEventListener
-     *
      * Remove a previously installed event handler
      */
     removeEventListener: function (eventName, handler) {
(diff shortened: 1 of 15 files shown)


src/remotestorage.js
```diff
@@ -682,7 +667,16 @@ RemoteStorage.prototype = {
...
+   * Start synchronization with remote storage, downloading and uploading any
+   * changes within the cached paths.
+   *
+   * Please consider: local changes will attempt sync immediately, and remote
+   * changes should also be synced timely when using library defaults. So
+   * this is mostly useful for letting users sync manually, when pressing a
+   * sync button for example. This might feel safer to them sometimes, esp.
+   * when shifting between offline and online a lot.
+   *
+   * TODO: move to sync.js
```
skddc: What about the TODO there? Didn't you just move the doc?

src/eventhandling.js
```diff
@@ -1,12 +1,10 @@
   var log = require('./log');
 
   /**
-   * Interface: eventhandling
+   * @interfacek
```
skddc: Typo with the k there?

src/indexeddb.js
```diff
@@ -90,6 +83,9 @@
   };
 
   IndexedDB.prototype = {
+    /**
+     * @todo Document
```
skddc: We use `TODO` comments instead of these currently.

skddc:
Awesome! Just added a couple of comments/questions.

src/indexeddb.js
gregkare: I'll switch them back to `TODO`

src/remotestorage.js
gregkare: I moved this comment that was prefixed with `startSync` from the top of the file to the top of the `startSync` function. Is `TODO: move to sync.js` saying that the comment needs to be moved to the `sync.sync` function?

skddc: I'm not sure what's unclear about the language there. But I don't see why it should be moved. I asked you what's with the TODO there, because you looked at the doc and apparently already moved it to the appropriate place.

gregkare:
I fixed the issues and removed the `TODO: move to sync.js` since it doesn't seem to make sense. I think this should be ready to merge. Do we want to generate docs for the other files or is that not needed?

skddc:
> Do we want to generate docs for the other files or is that not needed?

You mean for private APIs? I think we have all public functions covered now, or maybe something is missing?

PR #1021, opened 2017-10-14
Improve usage of setApiKeys

skddc:
Currently, one needs to call the same function twice in order to set api keys for both Dropbox and Google Drive. This changes the signature to accept (and require) a config object with backend type as keys and api keys as properties.  It also removes the need to pass in specific key objects with `appKey` and `clientId` properties.


galfert:
Nice one. Much better 👍

PR #1023, opened 2017-10-18
Add docs about AMD with RequireJS

gregkare:
Connected to #1003 

The module is exported as `RemoteStorage`, so I had to configure RequireJS to look for it in the `remotestorage.js` file.


skddc:
Squashed and merged locally in 290a78fbd79e1940a871a9042ff1556f1de25f5f

PR #1026, opened 2017-10-19
Transpile Array.includes

galfert:
`includes` is not available in e.g. IE 11.


skddc:
thumbs-up-asian-guy-gif

PR #1031, opened 2017-10-20
Don't throw error when deleting non-existent item

### LINKED ISSUE #993: Error when trying to delete last item in directory
As noted in https://github.com/remotestorage/remotestorage.js/pull/975#issuecomment-301040742

> When trying to delete the last bookmark, it gets deleted locally, but I can see the item still in my Dropbox. When disconnecting and reconnecting I also see the Bookmark in Webmarks again. But then when trying to delete it again, the remove call's promise fails with an error
>
>> Cannot delete non-existing node /

This seems to happen for both the Dropbox and the Google Drive backends.

skddc:
Instead just log what happened to the console.

fixes #993

### DIFF
--- a/src/cachinglayer.js
+++ b/src/cachinglayer.js
@@ -188,7 +188,8 @@
           var previous;
 
           if (!node) {
-            throw new Error('Cannot delete non-existing node '+path);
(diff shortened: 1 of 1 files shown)


galfert:
Good job :)

Issue #1033, opened 2017-10-23
Document all config options of the RemoteStorage constructor argument


skddc:

gregkare:
Closed by d9792998f040be8587f2f10878be3568b96cb39d

PR #1038, opened 2017-10-24
Document the RemoteStorage constructor's config argument

gregkare:
I'm not sure what the best format for this is, so far I used comments above each value and duplicated it in `src/config.js` as well.

https://github.com/erikrose/sphinx-js says this about abusing autofunction to document @typedef tags, but I found this to be confusing when I tried it, especially since our `config` is just a variable.

> There is experimental support for abusing autofunction to document @typedef tags as well, though the result will be styled as a function, and @property tags will fall misleadingly under an "Arguments" heading. Still, it's better than nothing until we can do it properly.

What do you think?

Connected to #1033

### DIFF
--- a/doc/js-api/remotestorage.rst
+++ b/doc/js-api/remotestorage.rst
@@ -15,15 +15,21 @@ The constructor can optionally be called with a configuration object, for
 example::
 
    var remoteStorage = new RemoteStorage({
+     // enable caching, defaults to true
      cache: true,
+     // Change Events that are enabled, default to true except for window
+     // which is false
      changeEvents: {
        local:    true,
        window:   false,
        remote:   true,
        conflict: true
      },
+     // set a redirect URI for Cordova apps, defaults to undefined
      cordovaRedirectUri: undefined,
+     // enable remoteStorage logging, defaults to false
      logging: false,
+     // extra remoteStorage data modules to load, defaults to an empty list
      modules: []
    });
 
--- a/doc/js-api/remotestorage.rst
+++ b/doc/js-api/remotestorage.rst
@@ -16,6 +16,43 @@ example::
      cordovaRedirectUri: 'https://app.wow-much-app.com' // defaults to undefined
    });
 
+Here is a complete list of the default options (set in ``src/config/js``), all
+of which can be overridden by calling the constructor with a configuration
+object like above::
+
+   {
+     // Enable remoteStorage logging
+     logging: false,
+     // Change Events that are enabled. See the BaseClient API for more information
+     changeEvents: {
+       local:    true,
+       window:   false,
+       remote:   true,
+       conflict: true
+     },
+     // enable caching
+     cache: true,
+     // disable specified modules
+     disableFeatures: [],
+     // timeout for the Webfinger lookup, discovering a connecting user's storage details
+     discoveryTimeout: 10000,
+     // sync interval when the application is in the foreground
+     syncInterval: 10000,
+     // sync interval when the application is in the background
+     backgroundSyncInterval: 60000,
+     // initial value for the internal state of the app being in the background
+     isBackground: false,
+     // set a redirect URI for Cordova apps
+     cordovaRedirectUri: undefined,
+     // timeout for network requests
+     requestTimeout: 30000,
+     // extra ES6 modules to load
+     modules: []
+   };
+
+For more information about the Change Events, see
+:doc:`the BaseClient API doc</js-api/base-client>`.
+
 .. NOTE::
    In the current version, it is only possible to use a single
    ``RemoteStorage`` instance. You cannot connect to two different remotes yet.
--- a/doc/js-api/remotestorage.rst
+++ b/doc/js-api/remotestorage.rst
@@ -23,7 +23,6 @@ example::
        conflict: true
      },
      cordovaRedirectUri: undefined,
-     disableFeatures: [],
      logging: false,
      modules: []
    });
--- a/doc/js-api/remotestorage.rst
+++ b/doc/js-api/remotestorage.rst
@@ -4,6 +4,9 @@ RemoteStorage
 Constructor
 -----------
 
+.. autofunction:: RemoteStorage([config])
+  :short-name:
+
 Create a ``remoteStorage`` instance like so::
 
    var remoteStorage = new RemoteStorage();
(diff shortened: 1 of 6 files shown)


doc/js-api/remotestorage.rst
```diff
@@ -16,6 +16,43 @@ example::
...
+       window:   false,
+       remote:   true,
+       conflict: true
+     },
+     // enable caching
+     cache: true,
+     // disable specified modules
+     disableFeatures: [],
+     // timeout for the Webfinger lookup, discovering a connecting user's storage details
+     discoveryTimeout: 10000,
```
skddc: Not public.

doc/js-api/remotestorage.rst
```diff
@@ -16,6 +16,43 @@ example::
...
+       local:    true,
+       window:   false,
+       remote:   true,
+       conflict: true
+     },
+     // enable caching
+     cache: true,
+     // disable specified modules
+     disableFeatures: [],
+     // timeout for the Webfinger lookup, discovering a connecting user's storage details
```
skddc: Not public.

doc/js-api/remotestorage.rst
```diff
@@ -16,6 +16,43 @@ example::
...
+     discoveryTimeout: 10000,
+     // sync interval when the application is in the foreground
+     syncInterval: 10000,
+     // sync interval when the application is in the background
+     backgroundSyncInterval: 60000,
+     // initial value for the internal state of the app being in the background
+     isBackground: false,
```
skddc: Not public.

doc/js-api/remotestorage.rst
```diff
@@ -16,6 +16,43 @@ example::
...
+     // set a redirect URI for Cordova apps
+     cordovaRedirectUri: undefined,
+     // timeout for network requests
+     requestTimeout: 30000,
+     // extra ES6 modules to load
+     modules: []
```
skddc: This is not about ES6 modules, but for loading remoteStorage data modules.

doc/js-api/remotestorage.rst
```diff
@@ -16,6 +16,43 @@ example::
...
+   };
+
+For more information about the Change Events, see
+:doc:`the BaseClient API doc</js-api/base-client>`.
```
skddc: This is documented elsewhere. No need to link to there from here, but important to link to here from there.

doc/js-api/remotestorage.rst
```diff
@@ -16,6 +16,43 @@ example::
      cordovaRedirectUri: 'https://app.wow-much-app.com' // defaults to undefined
    });
 
+Here is a complete list of the default options (set in ``src/config/js``), all
+of which can be overridden by calling the constructor with a configuration
+object like above::
```
skddc: Why not just extend the object above? It already has comments for the defaults (and much more concise by just having them in the same line with minimal text).

src/config.js
```diff
@@ -1,19 +1,34 @@
...
+  // Change Events that are enabled. See the BaseClient API for more information
   changeEvents: {
     local:    true,
     window:   false,
     remote:   true,
     conflict: true
   },
+  // enable caching
   cache: true,
+  // disable specified modules
```
skddc: Modules -> Features

skddc:
I'm not sure about the comments in `config.js`. This stuff should be documented from where it's actually set/get imo, because without context this information is rather useless, and the ones that don't need context have expressive names anyway.

If something is important here, it's probably which of these are supposed to be public and which aren't, and then later moving the ones that are not out of this config.

doc/js-api/remotestorage.rst
gregkare: I thought it was because of this comment: https://github.com/remotestorage/remotestorage.js/blob/b01fa56028e5bc18e4f7ef1782c045a77ccb3dc3/src/discover.js#L20-L21

skddc: I don't how that got there, and it's not part of any public docs, but why would one want to change this value?

gregkare:
Yes, the public ones were not clear to me. I agree that marking them will be a good first step! I'll do that

I'll also look into how to document them where they're set/get

skddc:
> I agree that marking them will be a good first step! I'll do that

I said we didn't have to:

> But without marking them as such, we already have that documented by way of the public ones appearing in the constructor doc.

gregkare:
I have moved the documentation of the public config options to the `RemoteStorage` constructor's API docs, I think this is much better now

gregkare:
Here's a screenshot:

screen shot 2017-10-25 at 13 22 06

doc/js-api/remotestorage.rst
```diff
@@ -12,46 +15,18 @@ The constructor can optionally be called with a configuration object, for
...
-     // sync interval when the application is in the background
-     backgroundSyncInterval: 60000,
-     // initial value for the internal state of the app being in the background
-     isBackground: false,
-     // set a redirect URI for Cordova apps
      cordovaRedirectUri: undefined,
-     // timeout for network requests
-     requestTimeout: 30000,
-     // extra ES6 modules to load
+     disableFeatures: [],
```
skddc: Still not public.

doc/js-api/remotestorage.rst
```diff
@@ -12,46 +15,18 @@ The constructor can optionally be called with a configuration object, for
...
-
-Here is a complete list of the default options (set in ``src/config/js``), all
-of which can be overridden by calling the constructor with a configuration
-object like above::
-
-   {
-     // Enable remoteStorage logging
-     logging: false,
-     // Change Events that are enabled. See the BaseClient API for more information
+     cache: true,
```
skddc: Where did the comments with the defaults go?

gregkare: I moved them to the constructor's argument docs, do you think it's better to have them both here and there?

doc/js-api/remotestorage.rst
gregkare: Oops, missed that one

gregkare:
@skddc I just saw that my comment answering "Where did the comments with the defaults go?" was hidden because I pushed, here is is again:

I moved them to the constructor's argument docs, do you think it's better to have them both here and there? If so I can add them back and this will be ready to merge

skddc:
I was looking at the argument docs and saw them being removed *there*. So perhaps GitHub is trolling me or buggy.

skddc:
screenshot from 2017-10-25 16-59-46

gregkare:
I have moved back the comments about the config options to the example:

screen shot 2017-10-25 at 17 18 19

skddc:
Fixed conflicts and merged locally.

PR #1039, opened 2017-10-24
Sync interval docs

gregkare:
I have reviewed the code examples from the new JS API docs (`RemoteStorage` class).

I have added a missing event as well as documentation about the allowed range of the sync interval.

I don't think I'm the right person to document `RemoteStorage#authorize()`, who has used this function before?

What do we want to do about the `enableLog` and `disableLog` functions? https://remotestoragejs.readthedocs.io/en/latest/js-api/remotestorage.html#enableLog

### DIFF
--- a/doc/js-api/remotestorage.rst
+++ b/doc/js-api/remotestorage.rst
@@ -86,6 +86,10 @@ List of events
    Emitted once when a wire request succeeds for the first time after a failed
    one, and ``remote.online`` is set back to true
 
+``sync-interval-change``
+"""""""""
+   Emitted when the sync interval changes
+
 Prototype functions
 -------------------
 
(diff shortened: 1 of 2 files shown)


skddc:
> What do we want to do about the enableLog and disableLog functions? https://remotestoragejs.readthedocs.io/en/latest/js-api/remotestorage.html#enableLog

Personally, I'm 100% for removing them, both because of the inherent race condition (logs never actually start at the beginning, when you use this function after initialization), as well as to only have a single correct way for enabling logs, instead of two different ways.

gregkare:
I have fixed the formatting error, this one can be merged

skddc:
Great!

PR #1040, opened 2017-10-24
Add code examples for all public RemoteStorage functions

gregkare:
I have also indicated functions that take optional arguments (`authorize` and `connect`)

### DIFF
--- a/doc/js-api/remotestorage.rst
+++ b/doc/js-api/remotestorage.rst
@@ -95,41 +95,89 @@ Prototype functions
 
 The following functions can be called on your ``remoteStorage`` instance:
 
-.. autofunction:: RemoteStorage#authorize(authURL, cordovaRedirectUri)
+.. autofunction:: RemoteStorage#authorize(authURL, [cordovaRedirectUri])
   :short-name:
 
-.. autofunction:: RemoteStorage#connect(userAddress, token)
+.. autofunction:: RemoteStorage#connect(userAddress, [token])
   :short-name:
+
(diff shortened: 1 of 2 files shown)


doc/js-api/remotestorage.rst
```diff
@@ -152,11 +200,29 @@ The following functions can be called on your ``remoteStorage`` instance:
...
+  Example::
+
+     remoteStorage.stopSync();
+
 .. autofunction:: RemoteStorage#onChange(path, handler)
   :short-name:
+
+  Example::
+
+     remoteStorage.onChange('/bookmarks/', function() {
```
skddc: This is fired for directories? So it's different than the BaseClient change events?

gregkare: Let me check

gregkare: Yes, a directory as a path works, I tried it in myfavoritedrinks. That's what I understood from the current docs. It also works with a path to a document. Should I document both cases? I had only tried with a directory before

skddc: Oh wait, I misread this. Directories are the normal use case here. But it needs to link to the extensive docs in BaseClient.

gregkare: Ack! Adding it

galfert: You might also want to document that the handler function gets passed the change object in the example.

skddc:
👍

Issue #1034, opened 2017-10-23
Missing XMLHttpRequest on node


gregkare:
Extracted from #1032 

```
$ node
> > var RemoteStorage = require('remotestoragejs');
... ReferenceError: XMLHttpRequest is not defined
...     at Object.<anonymous> (./node_modules/remotestoragejs/release/remotestorage.js:28:4332)
...     at t (./node_modules/remotestoragejs/release/remotestorage.js:2:352)
...     at Object.e.exports (./node_modules/remotestoragejs/release/remotestorage.js:28:53)
...     at t (./node_modules/remotestoragejs/release/remotestorage.js:2:352)
...     at Object.e.defaultValue (./node_modules/remotestoragejs/release/remotestorage.js:3:16231)
...     at t (./node_modules/remotestoragejs/release/remotestorage.js:2:352)
...     at Object.n (.node_modules/remotestoragejs/release/remotestorage.js:4:8757)
...     at t (./node_modules/remotestoragejs/release/remotestorage.js:2:352)
...     at Object.<anonymous> (./node_modules/remotestoragejs/release/remotestorage.js:2:472)
...     at t (./node_modules/remotestoragejs/release/remotestorage.js:2:352)
```

Including a dependency on a XMLHttpRequest library (such as xmlhttprequest) and instantiating a `XMLHttpRequest` variable before requiring `remotestorage` fixes it.

We used to require xhr2 as the XMLHttpRequest variable (c5d78f7d) and still have a dependency on it in package.json, but we're not using it in the code at the moment.

How should we fix this?

skddc:
I was trying to find out how to best handle this for about an hour, and I'm none the wiser now. The options I currently see:

### 1. Point the package's `main` back to an adapted `node-main.js` and use the not-entirely-but-almost-standard `browser` property for the browser build alone.

#### Pros:
* Node.js works out of the box by just requiring the module
* Only one package to know about and maintain

#### Cons:
* Browser builds might not work out of the box, but some (or most?) of them will (e.g. Browserify and Webpack would work afaics)

### 2. Create a second npm package, which is nothing but the current `node-main.js` and depends on `remotestoragejs`, could be `remotestoragejs-node`. Leave the UMD build as default in the main package and remove everything node-specific.

#### Pros:
* Node.js works out of the box with the node-specific package
* Browser packaging works out of the box with the normal/main package

#### Cons:
* Need to know about second package as user
* Need to maintain a second package (keep versions in sync and such)

[...]

PR #1043, opened 2017-10-24
Fix node build, configure browser build

skddc:
This seems to both fix rs.js when used with node.js, as well as still work with at least Browserify (tested with ember-browserify in Webmarks).

refs #1034
connected to #1034


skddc:
Here's a simple gist of what I used to make sure that the library loads correctly with node.js:

https://gist.github.com/skddc/9f3f5fc0c21b9d92db0d2bffad9a4c29

skddc:
I was playing around a bit more, and pushed the code I created: https://github.com/skddc/rsjs-fun

So, with Webpack it's obviously possible to just pack the whole thing from the node entry point, and because it doesn't export the `xhr2` module, but only sets it on the `global` for the duration of the node program execution, there's also no additional (node-specific) code in the resulting build.

The only potential downside I can see is that people don't add all the polyfills for things we might want to use in the source code, but that would be problematic or just a nuisance.

skddc:
This is ready to be considered for merge.

gregkare:
This works! I tested it on a hubot with the latest hubot-remotestorage-logger with the `global.XMLHttpRequest = require('xhr2');` line removed from `index.js` since it's done by in `node-main.js`

gregkare:
I have also tested it with plain Browserify

skddc:
We'll need some +1s for this to be merged.

gregkare:
I found the original commit adding the `WireClient.readBinaryData` fix: https://github.com/remotestorage/remotestorage.js/commit/5cc64669bff9b4324b74801ce3a68dffa1cf9039, it doesn't contain details though

skddc:
It will still need to overwrite this, if we're unable to detect browser environments here:

https://github.com/remotestorage/remotestorage.js/blob/master/src/wireclient.js#L100-L122

I'd rather detect the environment tbh.

galfert:
> Also, anyone knows what that WireClient.readBinaryData fix is about?

I can't remember exactly anymore, but I think it was to support older Browsers (e.g. Android).

skddc:
@galfert The fix for node is only to remove *all* browser-specific binary file handling. If you look at it, you'll see that it just pipes the `content` argument through to the callback without doing anything.

galfert:
Yes, I was talking about the function itself. In node the whole function isn't needed, so the "fix" just replaces it with a no-op function.

galfert:
Also still works fine when using no module system at all.

+1

skddc:
@galfert You merged this, but the binary data function would certainly fail in node.js now, no?

galfert:
> You merged this, but the binary data function would certainly fail in node.js now, no? I can't see an issue for that.

Yes, I'm currently working on that. I left the original node issue open, so I didn't think it was necessary to create a new one.

skddc:
Ok, I added a comment in the #1034 and re-assigned it from me to you.

Issue #732, opened 2014-06-23
As an app developer I want to know when syncing is done


xMartin:
When syncing documents or initially loading documents there's single events for each new or updated document. If I use these events to update the UI lots of updates are made in a short time causing things to flicker or jump around in case of updated order in a list or something and it is also just inefficient. I would like to know when a sync cycle or an initial loading is done so I can use that event to trigger one single update of the UI with all the new data.

skddc:
FYI: when `remoteStorage` is `ready`, there's `remoteStorage.sync.on('done')`. This will fire, when all single sync tasks have been completed and a new sync is queued. There's also `req-done` for the single tasks.

I think we should probably emit that event on the `remoteStorage` instance itself as well, in order for it to be a proper public event. Then again, maybe it's also nice to keep these things namespaced on the classes. Not sure. Any opinions?

skddc:
In 1.0.x, there is now an event `sync-req-done` and `sync-done` for both the single sync operations as well as a whole cycle completing (i.e. everything has been synced according to the caching config): https://remotestoragejs.readthedocs.io/en/latest/js-api/remotestorage.html#list-of-events

Also, the [`startSync()`](https://remotestoragejs.readthedocs.io/en/latest/js-api/remotestorage.html#startSync) function will return a promise, which resolves after a successful sync. However, it's recommended to use the events, as the library automatically syncs changes, both on app startup as well as periodically after that.

PR #1044, opened 2017-10-25
Move 'done' and 'req-done' events from sync to rs

galfert:
They are called 'sync-done' and 'sync-req-done' now.

refs #732

### DIFF
--- a/doc/js-api/remotestorage.rst
+++ b/doc/js-api/remotestorage.rst
@@ -70,19 +70,19 @@ List of events
 
 ``wire-busy``
 """""""""""""
-   Emitted when a wire request starts
+   Emitted when a network request starts
 
 ``wire-done``
 """""""""""""
-   Emitted when a wire request completes
+   Emitted when a network request completes
 
 ``sync-req-done``
 """""""""""""""""
-   Emitted when a single sync task has finished
+   Emitted when a sync request has finished
 
 ``sync-done``
 """""""""""""
-   Emitted when the complete sync cycle has finished
+   Emitted when a sync cycle has finished
 
 ``network-offline``
 """""""""""""""""""
--- a/doc/js-api/remotestorage.rst
+++ b/doc/js-api/remotestorage.rst
@@ -78,11 +78,12 @@ List of events
 
 ``sync-req-done``
 """""""""""""""""
-   Emitted when a sync request has finished
+   Emitted when a single sync request has finished
 
 ``sync-done``
 """""""""""""
-   Emitted when a sync cycle has finished
+   Emitted when all tasks of a sync have been completed and a new sync
+   is scheduled
 
 ``network-offline``
 """""""""""""""""""
--- a/doc/js-api/remotestorage.rst
+++ b/doc/js-api/remotestorage.rst
@@ -76,6 +76,14 @@ List of events
 """""""""""""
    Emitted when a wire request completes
 
+``sync-req-done``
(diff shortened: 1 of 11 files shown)


gregkare:
It's missing the documentation changes in the List of events (doc/js-api/remotestorage.rst)

galfert:
I added the documentation.

I noticed that the documentation also says that `remoteStorage` would emit the `wire-busy` and `wire-done` events as well. But they are emitted by `remoteStorage.remote` instead.

Should we move these events to `remoteStorage` as well?

skddc:
@galfert Are they needed as public API? We're not using them in the widget, right?

doc/js-api/remotestorage.rst
```diff
@@ -76,6 +76,14 @@ List of events
 """""""""""""
    Emitted when a wire request completes
 
+``sync-req-done``
+"""""""""""""""""
+   Emitted when a single sync task has finished
```
skddc: What does "sync task" mean exactly?

doc/js-api/remotestorage.rst
```diff
@@ -76,6 +76,14 @@ List of events
...
+
+``sync-done``
+"""""""""""""
+   Emitted when the complete sync cycle has finished
```
skddc: What's "the" here? As this document describes the events emitted by rs in general, what is "the complete sync cycle"? I think this is supposed to mean a/one sync cycle/process?

galfert:
Are they needed as public API? We're not using them in the widget, right?

They might be useful to use when caching is disabled for loading indication during up- and downloads.

skddc:
Ah, right. So let's move them?

doc/js-api/remotestorage.rst
galfert: If I understand correctly, a task can be a fetch of a directory or the upload of a local change.

skddc: So why not say "sync request" then? The event's name already uses it and seems to be correct and descriptive in this case.

skddc:
(FYI: I linked the old issue about knowing when sync is done in the PR description)

galfert:
Moved the wire events as well and wording of the documentation a bit.

doc/js-api/remotestorage.rst
```diff
@@ -70,19 +70,19 @@ List of events
...
 
 ``wire-done``
 """""""""""""
-   Emitted when a wire request completes
+   Emitted when a network request completes
 
 ``sync-req-done``
 """""""""""""""""
-   Emitted when a single sync task has finished
+   Emitted when a sync request has finished
```
skddc: I actually liked the "single" in there. Makes it absolutely clear what is meant.

doc/js-api/remotestorage.rst
```diff
@@ -70,19 +70,19 @@ List of events
...
 
 ``sync-done``
 """""""""""""
-   Emitted when the complete sync cycle has finished
+   Emitted when a sync cycle has finished
```
skddc: Again, not sure if removing "complete" is useful here, making it more ambiguous. I'd probably change "finished" to "completed" even, so it's clear when this is emitted. In a comment I found earlier, it also says "... and a new sync is scheduled", making it even clearer.

galfert: I like that description, will change it

skddc:
LGTM

Issue #1005, opened 2017-10-05
Move, edit, rewrite, delete the legacy docs


skddc:
They're all in `doc/legacy/` now. Most of them probably need entire rewrites, but some could probably survive with minor edits. If you pick one to work on, just add a comment, so other people don't grab the same one in parallel.

* [x] AMD (deleted)
* [ ] API
* [x] Caching
* [x] Data format
* [x] Discovery bootstrap
* [x] Node.js #1047

skddc:
Closing this, as the only things left in legacy are now modules, for which we have #1004

PR #1047, opened 2017-10-26
Rewrite the Node.js documentation

gregkare:
Connected to #1005


doc/nodejs.rst
```diff
@@ -0,0 +1,57 @@
...
+
+remoteStorage.js modules are not packaged as part of the ``remotestoragejs``
+npm module. You will need to download (or create) your own module to use. You
+can visit our `modules repository`_ to see existing modules, feel free to
+submit a pull request if you've written your own.
+
+They can also be published as npm modules: `search for "remotestorage-module" on npm`_ to find them.
+
+As of remoteStorage.js 1.0 you do not need to declare a ``remoteStorage`` global
+for modules to work in Node.js any longer, because they can be ES6 modules.
```
skddc: 1. Modules need to be documented for everyone, hence #1004 
2. I think information about no longer having to use a global is irrelevant for this document. If anything, it belongs in the CHANGELOG or announcement post imo.

doc/nodejs.rst
```diff
@@ -0,0 +1,57 @@
...
+
+Installation
+------------
+
+Client library
+^^^^^^^^^^^^^^
+
+remoteStorage.js is available from npm::
+
+   $ npm install remotestoragejs
```
skddc: This is already documented in https://remotestoragejs.readthedocs.io/en/latest/getting-started/how-to-add.html

doc/nodejs.rst
```diff
@@ -0,0 +1,57 @@
...
+
+Caveats
+-------
+
+* The OAuth browser redirect is not supported, you need to pass a token to the
+  `remotestorage.connect()` as the second argument. See the documentation on
+  the :doc:`RemoteStorage API doc</js-api/remotestorage>`. For example you
+  can use a web app to connect to the storage server and get the OAuth bearer
+  token using ``remoteStorage.remote.token`` in the console.
+* IndexedDB and LocalStorage are not supported in a Node.js environment, for
```
skddc: This fails to explain the actual caveat, namely that there is no persistent local cache between sessions/program executions.

doc/nodejs.rst
```diff
@@ -0,0 +1,57 @@
...
+Examples
+--------
+
+* `hubot-remotestorage-logger`_, a Hubot script that logs chat messages to
+  remoteStorage-enabled accounts using the `remotestorage-module-chat-messages`_
+  module
+* `rs-backup`_, a program that allows you to backup your data from a
+  remoteStorage account to a local hard drive and restore it to the same or
+  another account or server. It does not use remoteStorage modules or the
+  remoteStorage.js client library.
```
skddc: If it doesn't use this library, how is it relevant then? This needs to be linked directly from the section about having to acquire a token via other means, and ideally both include a description of how that can be done, and link directly to the code that does it.

doc/nodejs.rst
```diff
@@ -0,0 +1,57 @@
...
+  token using ``remoteStorage.remote.token`` in the console.
+* IndexedDB and LocalStorage are not supported in a Node.js environment, for
+  local storage only in-memory storage is supported.
+
+
+Examples
+--------
+
+* `hubot-remotestorage-logger`_, a Hubot script that logs chat messages to
+  remoteStorage-enabled accounts using the `remotestorage-module-chat-messages`_
```
skddc: `remotestorage-module` is just the prefix in module names. We would call it the "chat-messages module".

doc/nodejs.rst
```diff
@@ -0,0 +1,57 @@
...
+for modules to work in Node.js any longer, because they can be ES6 modules.
+
+Caveats
+-------
+
+* The OAuth browser redirect is not supported, you need to pass a token to the
+  `remotestorage.connect()` as the second argument. See the documentation on
+  the :doc:`RemoteStorage API doc</js-api/remotestorage>`. For example you
+  can use a web app to connect to the storage server and get the OAuth bearer
+  token using ``remoteStorage.remote.token`` in the console.
```
skddc: I think this should not merely be mentioned as a caveat emptor, but prominently at the top as the main difference between node.js and normal usage. Including the links to examples *in context* and with description of how/what they do and why.

skddc:
This is a good start! I added some comments/suggestions.

gregkare:
Thanks for the comments! On it

gregkare:
I pushed changes, can you take another look?

skddc:
Yup.

skddc:
... by the way, please use issue numbers for branch names, as explained in https://remotestoragejs.readthedocs.io/en/latest/contributing/github-flow.html#branch-names -- usually there's at least a tracking issue, like e.g. #1005 for this one. If there's no issue at the time of creating the branch, then it usually means you should create one first, or amend an existing one.

skddc:
I did quite some editing/writing on this one. Can someone have another look please?

doc/nodejs.rst
```diff
@@ -0,0 +1,80 @@
...
+For CLI programs, and if you actually want to integrate the OAuth flow in your
+program, one possible solution is the following:
+
+1. Set up a simple Web site/app, which you publish under a fitting domain/URI
+   that you can use as the OAuth redirect URI.
+2. Have the user enter their user address and do a Webfinger lookup for auth
+   URL etc., e.g. using `webfinger.js
+   <https://www.npmjs.com/package/webfinger.js>`_.
+3. Create the OAuth request URI with the correct scope etc., and open a browser
+   window with that URI from your program.
```
galfert: Maybe add "or prompt the user to open it".

galfert:
Nice work. Only made one minor comment, but that's not really important. Please decide if you want to add it or not and just merge yourself then.

doc/nodejs.rst
skddc: Makes sense. Thanks.

PR #1049, opened 2017-10-27
Fix Dropbox binary uploads

### LINKED ISSUE #1048: Error when uploading binary files to Dropbox
Looks like the client is using a wrong content-type when uploading. The error response from Dropbox is:

> Error in call to API function "files/upload": Bad HTTP "Content-Type" header: "application/json; charset=UTF-8".  Expecting one of "application/octet-stream", "text/plain; charset=dropbox-cors-hack".

galfert:
Fixes #1048 

This fixes Dropbox converting binary data to JSON.

It also removes the use of `WireClient.readBinaryData`, which has been obsolete for quite a while now.

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,9 +2,4 @@ global.XMLHttpRequest = require('xhr2');
 
 const RemoteStorage = require('./src/remotestorage.js');
 
-// TODO re-introduce fix (or verify it's obsolete)
-// RemoteStorage.WireClient.readBinaryData = function (content, mimeType, callback) {
-//   callback(content);
-// };
-
 module.exports = RemoteStorage;
--- a/index.js
+++ b/index.js
@@ -2,4 +2,9 @@ global.XMLHttpRequest = require('xhr2');
 
 const RemoteStorage = require('./src/remotestorage.js');
 
+// TODO re-introduce fix (or verify it's obsolete)
+// RemoteStorage.WireClient.readBinaryData = function (content, mimeType, callback) {
+//   callback(content);
+// };
+
 module.exports = RemoteStorage;
--- a/src/dropbox.js
+++ b/src/dropbox.js
@@ -432,11 +432,16 @@
 
         // handling binary
         if (!mime || mime.match(/charset=binary/)) {
-          return Promise.resolve({
-            statusCode: status,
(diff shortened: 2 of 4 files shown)


galfert:
When I tested the removal of `readBinaryData` I just checked that the response was the same before and after removing it.

But now I just noticed that the Dropbox client just doesn't detect binary files properly when downloading them. The problem is that Dropbox removes the charset from the content-type, so our check (see https://github.com/remotestorage/remotestorage.js/blob/81522690ec5608af6ab1fd20b40db78834682351/src/dropbox.js#L434) doesn't get triggered.

galfert:
Since I can't really confirm yet that the `readBinaryData` function is not needed for Dropbox anymore because of the content-type problem, I reverted the commit that removed it.
Instead I moved the function to the Dropbox client itself and made it return a Promise instead of using the callback. This way I could at least get rid of the FIXME comment :)

galfert:
> Why don't we need to load binary data into array buffers for RS servers as well?

The WireClient always loads data as array buffer:(https://github.com/remotestorage/remotestorage.js/blob/4cfea9b466554b3195d4233253175259bca872fb/src/wireclient.js#L269

and then extracts the text from it if it's not binary data: https://github.com/remotestorage/remotestorage.js/blob/4cfea9b466554b3195d4233253175259bca872fb/src/wireclient.js#L313

skddc:
Aha, thanks.

skddc:
@galfert Is this ready for merge now? If so, would probably be a good idea to explain this to someone again, as I personally e.g. don't even know what exactly it's fixing and not fixing now.

galfert:
This fixes only the references issue, setting the correct content-type when uploading binary files. For the other bug there is a separate issue (#1059).

The other commit in this PR is just a refactoring that shouldn't change anything.

skddc:
So can you please confirm "but downloads don't due to the reading issue?". And if that's the case, don't you think we should tell people?

galfert:
Yes, downloads don't work due to the reading issue. The documentation already mentions that Dropbox doesn't support content-types in the known issue. But I think we should add the implications for binary files to that.

But I wanted to check first if there might be a way to fix it first.

PR #1050, opened 2017-10-27
Extend "why use this" doc

skddc:
More reasons for using rs.js. Hope it doesn't sound too braggy.


gregkare:
I think it's good! LGTM

galfert:
Just the right amount of brag :)

Issue #1054, opened 2017-10-27
Implement getItemURL for Google Drive


galfert:
At the moment, when connected to Google Drive, it still used the BaseClient's method, which just gives a string in the form of `undefined/relative/path/to/file`.

PR #1055, opened 2017-10-27
Throw error about Google Drive getItemURL not being implemented

galfert:
Refs #1054

Until it's implemented, this throws an error when calling the function instead of just returning a wrong value.

### DIFF
--- a/src/baseclient.js
+++ b/src/baseclient.js
@@ -283,10 +283,6 @@ BaseClient.prototype = {
    * Retrieve full URL of a document. Useful for example for sharing the public
    * URL of an item in the ``/public`` folder.
    *
-   * Please note: This method currently only works for remoteStorage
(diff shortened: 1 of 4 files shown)


src/baseclient.js
```diff
@@ -283,6 +283,10 @@ BaseClient.prototype = {
    * Retrieve full URL of a document. Useful for example for sharing the public
    * URL of an item in the ``/public`` folder.
    *
+   * Please note: This method currently only works for remoteStorage
```
skddc: reStructuredText has all kinds of directives for additional information, e.g. HINT, TIP, WARN, among others, which are then rendered nicely in the output. I'd go with a warning here.

galfert: Thanks for the HINT :)

skddc:
lgtm

PR #1060, opened 2017-10-29
Add missing revision on successful put to Dropbox

### LINKED ISSUE #1042: Remote files not deleted in Dropbox
While working an a fix for #1035 I noticed that when creating a document and then deleting it in the same session (without reloading the Browser in between), the remote document doesn't get deleted.

Looks like it's because the local node doesn't have a `common.revision`, so the library thinks there is no remote object yet and thus doesn't attempt the DELETE request.

galfert:
Fixes #1042

This fixes the problem, that remote files would not be deleted when they were created in the same session.

### DIFF
--- a/src/dropbox.js
+++ b/src/dropbox.js
@@ -915,7 +915,8 @@
         }
 
         this._revCache.propagateSet(params.path, body.rev);
-        return Promise.resolve({ statusCode: response.status });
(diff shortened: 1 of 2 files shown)


skddc:
Works exactly as advertised. :)

Issue #1056, opened 2017-10-27
Clean up old/obsolete files in repo


skddc:
E.g. there's still the "generator" and some other things. We need to just go through all files and make sure they still make sense, and if they do, make sure they're either documented or that an issue exists for that.

PR #1062, opened 2017-10-29
Clean up repo

skddc:
Remove some obsolete things, rename test script to be more expressive

connected to #1056


package.json
```diff
@@ -12,7 +12,7 @@
   },
   "homepage": "https://remotestorage.io",
   "scripts": {
-    "test": "./scripts/test",
+    "test": "bash scripts/test-all.sh",
```
galfert: Why the `bash` prefix? What about people using a different shell?

galfert:
Looks all good except for the comment I made.

package.json
skddc: That's the reason, so that it doesn't break when they use a different shell. The script is written for bash. Also, so that it's entirely clear what is being run without looking into the script first.

galfert: Ah, makes sense.

gregkare: The script is executable and specifies `/bin/bash` in the shebang so you don't need to specify a shell, but it doesn't hurt. In this case we could have used `/bin/sh` because our script doesn't contain bashisms, I forgot to check when I extracted it from the Makefile

skddc: I think it's a good idea to just go with just one as baseline, and bash is common enough to be that imo.

PR #1065, opened 2017-10-30
New module docs

skddc:
connected to #1004


doc/data-modules/defining-data-types.rst
```diff
@@ -0,0 +1,84 @@
...
+         },
+         "url": {
+           "type": "string",
+           "format": "uri"
+         },
+         "tags": {
+           "type": "array",
+           "default": []
+         },
+       }),
```
gregkare: This parenthesis shouldn't be here (it's closed on line 33)

doc/data-modules/defining-data-types.rst
```diff
@@ -0,0 +1,84 @@
...
+   // and in your app:
+
+   remoteStorage.bookmarks.add({
+     title: 'Unhosted Web Apps',
+     url: 'https://unhosted.org',
+     tags: ['unhosted', 'remotestorage', 'offline-first']
+   })
+   .then(() => {
+     console.log('stored bookmark successfully');
+   }
```
gregkare: Should be:

```js
.then(() => {
  console.log('stored bookmark successfully');
})
.catch((err) => {
  console.error('validation error:', err);
});
```

(two parenthesis issues)

doc/data-modules/defining-data-types.rst
```diff
@@ -0,0 +1,84 @@
...
+errors in case the data format doesn't match::
+
+   var Bookmarks = { name: 'bookmarks', builder: function(privateClient, publicClient) {
+     // ...
+
+     return {
+       exports: {
+
+         add: function (bookmark) {
+           bookmark.id = md5Hash(bookmark.url); // hash URL for nice ID
```
gregkare: This is the only part of the example that will not work out of the box by copying and pasting (since the `md5Hash` function isn't defined). I think including a short hashing function or saying in the comment that this is not expected to be working code would be better. What do you think?

gregkare:
I went through these new docs, looks great to me except for parentheses issues that I fixed and one question about an exemple

galfert:
I really like the intro text. The rest also looks good to me.

Is this still WIP or why is there an empty "Utility modules" doc?

doc/data-modules/defining-data-types.rst
skddc: I never thought anyone would try this as working code. In fact, that's a bad idea, as it would store data right next to actual Webmarks data, which has different requirements.

I don't know what to do about it, but it's actually good that this is failing as is.

gregkare: Good point, I only tried it as part of the review to check the syntax. Since it's not even a complete snippet I actually think it's good like this

skddc:
It's not. The Utility Modules doc is a placeholder and not linked from anywhere currently, so we don't forget about it.

doc/data-modules/defining-data-types.rst
skddc: Yer, the next thing that would fail is `storeObject`, because the type is not defined in the same snippet.

skddc:
Can we merge this, or does someone still have objections?

gregkare:
I vote for merging!

PR #1066, opened 2017-10-31
Fix delayed event handler registration

galfert:
This should fix the issue reported in remotestorage/remotestorage-widget#56.

When registering event handlers after the events already fired, they would not be triggered.
In the widget issue [I commented](https://github.com/remotestorage/remotestorage-widget/issues/56#issuecomment-338649163) that although the `connected` event handler would not be triggered, at least the `ready` event would always get triggered. But that was actually not the case. It was only triggered, when the user was connected.

I fixed this so the `ready` is always triggered. And I also added checks for `connected` and `not-connected`, so they also get triggered when registering a handler after those events were originally fired already.

I also added documentation for the `on` function. I had to document the function in the `remotestorage.rst` document directly, because the function is created in the constructor and Sphinx wouldn't find the function and the associated JSDoc.

### DIFF
--- a/src/remotestorage.js
+++ b/src/remotestorage.js
@@ -106,11 +106,31 @@ var RemoteStorage = function (cfg) {
    * @private
    */
   this.on = function (eventName, handler) {
-    if (eventName === 'ready' && this.remote && this.remote.connected && this._allLoaded) {
-      setTimeout(handler, 0);
-    } else if (eventName === 'features-loaded' && this._allLoaded) {
-      setTimeout(handler, 0);
+    // check if the handler should be called immedtiately, because the
+    // event has happened already
+    switch(eventName) {
+      case 'features-loaded':
+        if (this._allLoaded) {
+          setTimeout(handler, 0);
+        }
+        break;
+      case 'ready':
+        if (this._allLoaded && this.remote) {
+          setTimeout(handler, 0);
+        }
+        break;
+      case 'connected':
+        if (this._allLoaded && this.remote && this.remote.connected) {
+          setTimeout(handler, 0);
+        }
+        break;
+      case 'not-connected':
+        if (this._allLoaded && this.remote && !this.remote.connected) {
+          setTimeout(handler, 0);
+        }
+        break;
     }
+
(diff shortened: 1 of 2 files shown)


src/remotestorage.js
```diff
@@ -98,19 +98,41 @@ var RemoteStorage = function (cfg) {
...
-   * @private
+   * @param {string} eventName - Name of the event
+   * @param {function} handler - Event handler
    */
   this.on = function (eventName, handler) {
-    if (eventName === 'ready' && this.remote && this.remote.connected && this._allLoaded) {
-      setTimeout(handler, 0);
-    } else if (eventName === 'features-loaded' && this._allLoaded) {
-      setTimeout(handler, 0);
+    // check if the handler should be called immedtiately, because the
```
gregkare: typo: s/immedtiately/immediately/

gregkare:
I have tried this, this fixes a new issue that I had been getting on my local myfavoritedrinks (only on Firefox), where I was getting a blank widget on initial page load for a few seconds

I have pushed a small fix for the docs, using `autofunction` to avoid writing the same docs twice

skddc:
> this fixes a new issue that I had been getting on my local myfavoritedrinks (only on Firefox), where I was getting a blank widget on initial page load for a few seconds

That's the issue this is supposed to fix. So that's good. :)

> I have pushed a small fix for the docs, using autofunction to avoid writing the same docs twice

Great!

skddc:
This is actually missing a changelog entry, but I only noticed after merging.

galfert:
> I have pushed a small fix for the docs, using autofunction to avoid writing the same docs twice

Did you actually try that it works? I'm pretty sure I did it the same way and got an error from Sphinx:

> No JSDoc documentation was found for object "RemoteStorage#on" or any path ending with that.

gregkare:
@galfert Yes, I built the docs locally before pushing

galfert:
Ok, weird. I spent quite some time yesterday trying to fix this issue. Then I gave up and duplicated the doc ¯\\_(ツ)_/¯

PR #1070, opened 2017-11-01
Adjust CONTRIBUTING.md for new docs

skddc:


gregkare:
LGTM! Should I merge this?

skddc:
The two-plus-ones-rule is currently suspended, until after the 1.0.0 release, so we can move faster.

PR #1073, opened 2017-11-01
Document reading/writing data

skddc:
connected to #1003

### DIFF
--- a/doc/getting-started/read-and-write-data.rst
+++ b/doc/getting-started/read-and-write-data.rst
@@ -1,0 +1,58 @@
+Reading and writing data
+========================
+
+As soon as your :doc:`RemoteStorage </js-api/remotestorage>` instance is ready
+for action (signaled by the ``ready`` event), we can start reading and writing
+data.
+
+Anonymous mode
+--------------
+
+One of the unique features of rs.js is that users are not required to have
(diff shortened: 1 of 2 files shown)


doc/getting-started/read-and-write-data.rst
```diff
@@ -0,0 +1,58 @@
...
+development. Using :func:`scope`, you can create a new BaseClient scoped to a
+given path::
+
+   const client = remoteStorage.scope('/foo/');
+
+   // List all items in the "foo/" category/folder
+   client.getListing('')
+     .then(listing => console.log(listing));
+
+   // Write some text to "foo/bar.txt"
```
galfert: The example writes to `quotes.txt`, not `bar.txt`.

galfert:
Nice.

Not sure, but maybe it should be noted that for the example to work you also need to claim access to the `foo` category?
Just thinking of people that didn't read the whole "Getting started" guide from the beginning but went directly to this document.

doc/getting-started/read-and-write-data.rst
skddc: Good catch, I was changing names back and forth at some point and missed a line.

skddc:
Wouldn't this warn you right in the console?

galfert:
Ah, right. It does.

PR #1074, opened 2017-11-01
Update Webpack and Babel loader

skddc:
This should give us a nice green badge from https://david-dm.org/remotestorage/remotestorage.js again.


galfert:
👍

PR #1078, opened 2017-11-02
Fix failing test

galfert:
Test was failing due to change in d6dc4a22.

Please don't merge yet, I'll try to fix Travis testing PRs with this as well.

PR #1079, opened 2017-11-02
Fix code indentation

galfert:
I just wanted to create a new PR to test if Travis testing is properly triggered now. But I thought I could at least make a change that somewhat makes sense :)

This removes the two leading spaces that were left when removing the outer function scope during the ES6 refactoring.

Feel free to just close this if you think this is too much of a change right now.

### DIFF
--- a/src/access.js
+++ b/src/access.js
@@ -1,176 +1,176 @@
+/**
+ * @class Access
+ *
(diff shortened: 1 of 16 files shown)


galfert:
Yay, Travis did its job. The failing test is the one that's fixed with #1078.

PR #1082, opened 2017-11-02
Fix sync not being scheduled

### LINKED ISSUE #1081: Periodic sync broken since RC2
We forgot to actually use the new event for rescheduling: https://github.com/remotestorage/remotestorage.js/blob/master/src/remotestorage.js#L708-L717

skddc:
It was still using the old event.

fixes #1081

### DIFF
--- a/src/remotestorage.js
+++ b/src/remotestorage.js
@@ -705,7 +705,7 @@ RemoteStorage.prototype = {
       return;
     }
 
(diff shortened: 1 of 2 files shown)


skddc:
Rather critical bug. Please review asap!

galfert:
Damn, looks like I missed that one when I moved the events. Sorry.

Fix looks good to me.

PR #1083, opened 2017-11-03
Configure Codeclimate to do complexity checks

galfert:


skddc:
Looks like the Travis command is invalid:

```
Invalid option '--quit' - perhaps you meant '--quiet'?
```

PR #1085, opened 2017-11-04
Remove UUID and MD5 utils/deps

### LINKED ISSUE #1084: node.js execution fails with: "can't find module "uuid/v4"

skddc:
These functions are not used in this library. Any module or package, which depends on these hashing functions should require appropriate npm modules itself.

In the case of the UUID dep, this actually broke node.js in production, as the module is required in the code, but actually listed as dev dependency (instead of normal dependency).

fixes #1084


galfert:
👍

PR #1087, opened 2017-11-05
Fix disconnect in Safari

### LINKED ISSUE #1086: Can't fully disconnect in Safari
When trying to disconnect, the `disconnected` event never gets fired in Safari. So the widget stays showing as connected, although the storage is actually disconnected and there are no more syncs happening.

galfert:
Fixes #1086 

This doesn't wait for the IndexedDB to be deleted before resolving the `_rs_cleanup()` promise, so the disconnected event gets fired in Safari.

### DIFF
--- a/src/indexeddb.js
+++ b/src/indexeddb.js
@@ -441,8 +441,9 @@ IndexedDB._rs_cleanup = function (remoteStorage) {
       remoteStorage.local.closeDB();
     }
 
-    IndexedDB.clean(DEFAULT_DB_NAME, resolve);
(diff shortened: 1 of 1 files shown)


skddc:
I don't see anything about why we wouldn't want to wait for the database to be deleted. If we don't, it might still contain data from a previous session/user. Especially if this works in other browsers.

If this is a Safari bug, then please include documentation on it in this pull request. Also, we should then probably try to only break the normal behaviour in Safari, and not in browsers in which the callback works.

galfert:
The whole bug was invalid. Was just my IndexedDB being broken.

PR #1096, opened 2017-11-11
Fix Dropbox throwing an error on failed requests

galfert:
As reported in remotestorage/remotestorage-widget#69, the Dropbox client throws an error when offline.

This catches the error and just ignores it, because the offline handling (e.g. emitting the appropriate events) is already handled.

### DIFF
--- a/src/dropbox.js
+++ b/src/dropbox.js
@@ -796,6 +796,13 @@ Dropbox.prototype = {
         if (responseBody.has_more) {
           return fetch(responseBody.cursor);
         }
+      }).catch((error) => {
+        if (error instanceof ProgressEvent) {
+          // Offline is handled elsewhere already, just ignore it here
+          return Promise.resolve();
+        } else {
+          return Promise.reject(error);
+        }
       });
     };
 


skddc:
Unfortunately, the commit message doesn't explain what's actually happening and how it's being fixed here. I tried to understand it and read through WireClient and request code, but I'm still a bit lost about why this is only happening in this exact place and subsequently only being ignored there as a fix.

galfert:
To be honest, I don't fully understand it either. But when testing, this was the only place it was throwing an error.

skddc:
So you tested putting a file while offline in other backends? The reason you might only have seen it with Dropbox is that some of its sync requests actually do POSTs afaics.

It's just using `WireClient.request` really, so the same request type should be failing like that in all backends that don't handle the case.

galfert:
Yes, I tested all three backends, creating, editing and deleting content. This one was the only error that wasn't caught properly.

skddc:
Strange, but ok. Let's merge it for now, and refactor the requests when we get to it.

PR #1098, opened 2017-11-13
Link to remotestorage.io from docs

jorinvo:
I had the experience of ending up on the doc page while I was actually trying to go to the main website and I couldn't find any link.
Would be good to add some reference. The phrasing can be changed though.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -9,8 +9,10 @@ remoteStorage.js
 Welcome to the remoteStorage.js documentation!
 
 remoteStorage.js is a JavaScript library for storing user data locally in the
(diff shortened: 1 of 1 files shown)


skddc:
Good idea. Perhaps we should just replace the intro paragraph entirely with the new one from the [README](https://github.com/remotestorage/remotestorage.js/blob/master/README.md)?

jorinvo:
sure, didn't look at the readme. I updated the PR.

skddc:
Cool. Thanks!

skddc:
Dang, just realized that this is RST, not Markdown. Will fix in master.

Issue #1094, opened 2017-11-09
Google and Dropbox remote storage backends should be opt-in


rchrd2:
I notice they are being included and initialized by default. In many cases, these might not be used, so I would be ideal if they were optional. 

On a high level, here are some ideas:

In the dist folder they could be separate fles: eg `remotestorage-google.js`. 
For `import` it could be something like `import GoogleStorage from 'remotestoragejs-google';

Here is a screenshot showing that they are included with the standard `import RemoteStorage from 'remotestoragejs';`.

![screen shot 2017-11-08 at 10 53 31 pm](https://user-images.githubusercontent.com/182479/32592270-d80ac66c-c4d7-11e7-81f6-c9bca31391c8.png)

skddc:
The back-ends *are* actually opt-in. The feature is built into the code, but it will only be used when you configure OAuth app keys via [`RemoteStorage#setApiKeys`](https://remotestoragejs.readthedocs.io/en/latest/js-api/remotestorage.html#setApiKeys).

Also, if these are not set, then the [connect widget](https://github.com/remotestorage/remotestorage-widget) will not show any options to choose from and instead open the connect form for remoteStorage accounts directly. You can see an example of that on https://sharesome.5apps.com.

How it works in the code: `remoteStorage.remote` will be initialized as one of `WireClient`, `Dropbox` or `GoogleDrive`. So basically the two non-standard options are mostly different implementations of `WireClient`.

Hope this helps.

rchrd2:
Yes. My proposal is to remove Dropbox and GoogleDrive from being included in the core codebase. As it is now, if you aren't using them, the clients are still included in RemoteStorageJS.

Additionaly, I think that if they were separate modules that could be optionally included, it would also make it easier for others to create their own remote storage option. All the necessary hook-ins would be be setup as a consequence of keeping Dropbox and Google separate.

I understand that this might not be the highest priority, but I wanted to point it out.

Thanks for your consideration.

skddc:
I see. They're not actually taking up enough space for that to be a drawback currently. But I think it would be nice to make them seperate modules and with that create a proper API for additional back-ends. As you say, there are currently more important areas to focus on, but let's keep it open so potential contributors can find it.

I changed the title of the issue to reflect the nature of the proposal a little bit better. Think it's OK like this?

PR #1099, opened 2017-11-14
Use cordovaRedirectUri on Cordova

### LINKED ISSUE #1097: cordovaRedirectUri not working
Hello there

I am experiencing an issue with Cordova: it doesn't use the cordovaRedirectUri and instead uses the file:// path of the app (resulting in an error 400).

I have tried using
```ts
let remoteStorage = new RemoteStorage({
    ...
    cordovaRedirectUri: 'https://my.app.co.za'
});
```
And 
```ts
remoteStorage.setCordovaRedirectUri('https://my.app.co.za');
```
Where "my.app.co.za" is my own server's subdomain.

All code is running after cordova platform is ready, and window.cordova definitely exists. I have the in-app browser plugin installed. I'm currently running cordova@6.2.3.

The way I am connecting DropBox is as follows:
```ts
remoteStorage.dropbox.connect();
```

Any suggestions? Thanks

MrPutuLips:
Use config.cordovaRedirectUri when the platform is Cordova.

If the user has not set the cordovaRedirectUri it isn't used.

fixes #1097

### DIFF
--- a/src/authorize.js
+++ b/src/authorize.js
@@ -1,5 +1,6 @@
 var log = require('./log');
 var util = require('./util');
+var config = require('./config');
 
 function extractParams(url) {
   //FF already decodes the URL fragment in document.location.hash, so use this instead:
@@ -38,6 +39,13 @@ function extractParams(url) {
 
 
 var Authorize = function (remoteStorage, authURL, scope, redirectUri, clientId) {
+  // if the platform is cordova, change redirectUri's value to cordovaRedirectUri.
+  // this will only happen if the user has set cordovaRedirectUri.
+  if (util.globalContext.cordova &&
+    typeof config.cordovaRedirectUri !== 'undefined') {
+    redirectUri = config.cordovaRedirectUri;
+  }
+
   log('[Authorize] authURL = ', authURL, 'scope = ', scope, 'redirectUri = ', redirectUri, 'clientId = ', clientId);
 
   // keep track of the discovery data during redirect if we can't save it in localStorage


galfert:
Great. Thanks!

Code looks good to me. Will test it before merging.

skddc:
Please don't merge without a test. Thank you.

skddc:
I have interesting news on this one, discovered while trying to write a test for this change:

When connecting an RS account, the Cordova redirect URI is set in the public [`authorize()`](https://remotestoragejs.readthedocs.io/en/latest/js-api/remotestorage.html?highlight=authorize#authorize) function ([source code](https://github.com/remotestorage/remotestorage.js/blob/master/src/remotestorage.js#L174-L189)), which ironically is the only public one that has yet to be documented in that class.

And it turns out, the Dropbox and Google Drive backends are not using the public API, but directly call the private API. However, the public function doesn't yet allow to add a custom client ID, and it's receiving the Cordova redirect URI as an argument.

Hence, the correct fix is to extend the public function, and then use it from everywhere (and of course document it on the way). The great thing about this approach is that it is also another step toward offering a public storage adapter API, as for example requested in #1094.

It's after midnight here, so I'll leave it untouched for now, and only create a small PR with some Authorize tests that I improved/created on the way.

galfert:
When I tried testing this PR, I found that with Dropbox the redirect to the Dropbox OAuth works now, but on the redirect back, the OAuth token is not extracted properly. I couldn't find out yet, why that is.

Unfortunately for Google Drive, the redirect didn't work at all. Instead I got this error:

img_8332

skddc:
> with Dropbox the redirect to the Dropbox OAuth works now, but on the redirect back

Just to clarify: it's not doing a redirect to Dropbox, but opening a new window with the OAuth URL as argument. It then redirects to the redirect URI from the respective OAuth URI's param.

Regarding Google: looks like we can't support it for now then. There's likely some kind of Cordova plugin one could use however, and extending the public function as I explained can't hurt for somehow supporting that instead.

galfert:
> Hence, the correct fix is to extend the public function, make it use the config directly, and then use it from Dropbox and GDrive as well.

FYI: I started with this last night and I'm almost finished. Will create a PR later today.

skddc:
@galfert Please use Waffle and/or the tags to change the status *before you take sth over*. I almost started this as well, and we wouldn't doubled the effort.

galfert:
Btw, I managed to solve the Google Drive "disallowed_useragent" error by setting a different UserAgent string. You just need to put something like this in Cordova's `config.xml`:

```xml
<preference name="OverrideUserAgent" value="Mozilla/5.0 remoteStorage" />
```

skddc:
Cool. That should go in the Cordova docs, of course.

galfert:
I added it as a TODO to my doc PR (#1088).

skddc:
Closing in favor of #1101

@MrPutuLips Thank you so much for contributions on this one! It would be super helpful, if you could also verify @galfert's new fix with your app, and if you like also check out how the new code works. 🙏

PR #1101, opened 2017-11-23
Fix Google Drive and  Dropbox authorization in Cordova apps

### LINKED ISSUE #1097: cordovaRedirectUri not working

galfert:
Fixes #1097 

This changes the `remoteStorage.authorize()` method to allow to set all the possible options and use the `cordovaRedirectURL` when in a Cordova app.

The Google Drive and Dropbox clients now use this method instead of `Authorize()`.

With this, the authorization for these two backends works in Cordova as well now. But when testing I saw some weird behaviour with my test app. After authorization the app would often times just stop responding. Maybe someone else can test it with a different app.


src/authorize.js
```diff
@@ -36,6 +36,22 @@ function extractParams(url) {
...
+  url += '&scope=' + encodeURIComponent(scope);
+  url += '&client_id=' + encodeURIComponent(clientId);
+
+  if (hashPos !== - 1 && hashPos+1 !== redirectUri.length) {
+    url += '&state=' + encodeURIComponent(redirectUri.substring(hashPos+1));
+  }
+  url += '&response_type=token';
+
+  return url;
+}
```
skddc: I can't see a test for this function in the commit.

src/remotestorage.js
```diff
@@ -172,20 +172,31 @@ RemoteStorage.prototype = {
   },
 
   /**
-   * TODO: document
+   * Initiate the OAuth authorization flow.
```
skddc: This should contain a description of why/when/in what context one might want to use the function.

src/remotestorage.js
```diff
@@ -172,20 +172,31 @@ RemoteStorage.prototype = {
...
     this.access.setStorageType(this.remote.storageType);
-    var scope = this.access.scopeParameter;
+    if (typeof options.scope === 'undefined') {
+      options.scope = this.access.scopeParameter;
+    }
 
-    var redirectUri = globalContext.cordova ? cordovaRedirectUri : String(Authorize.getLocation());
+    if (typeof options.redirectUri === 'undefined') {
+      options.redirectUri = globalContext.cordova ? config.cordovaRedirectUri : String(Authorize.getLocation());
+    }
```
skddc: This changes the logic, and I think in a bad way. It's valid to set the redirect URI in an app that is both used in Cordova and outside of it. The original `if` condition was correct in that sense, but this one would not set the Cordova redirect URI, if the normal one is set.

Ironically, among the new tests, there's no test for setting the Cordova redirect URI.

src/authorize.js
galfert: I only extracted this to a (private) function from `Authorize()`. Everything is already tested implicitly by the `Authorize()` tests.

In fact you can change any of the lines in this method and will result in a test failure.

src/remotestorage.js
```diff
@@ -172,20 +172,31 @@ RemoteStorage.prototype = {
   },
 
   /**
-   * TODO: document
+   * Initiate the OAuth authorization flow.
    *
-   * @param {string} authUrl
-   * @param {string} cordovaRedirectUri
+   * @param {object} options
+   * @param {string} [options.authURL] - URL to the authorization endpoint
```
skddc: s/to/of

src/authorize.js
skddc: Yes, but as it's not part of the tested function anymore, this connection is not clear to anyone who's not you (which even includes future-you).

src/remotestorage.js
galfert: I changed it and added the test.

src/remotestorage.js
galfert: Not sure what to put there. Any suggestions?

src/authorize.js
galfert: It's a private function and you are not supposed to test private functions. I can't even call the function from the tests.

src/remotestorage.js
skddc: You just implemented it, so you know best about when to use it.

(The way I understand it, this is not usually something one wants to use in an app, but only in a custom storage backend. But I'm not 100% sure.)

src/authorize.js
skddc: Ok, makes sense. Thanks for explaining.

skddc:
@MrPutuLips Could you check out this fix, so we know it works for you as well, before releasing a new version?

skddc:
@galfert What exactly does "After authorization the app would often times just stop responding" mean? Could you describe it a bit more? Does it happen on both iOS and Android? Could you post an ADB log from the process?

galfert:
It just wouldn't take any user interaction anymore (e.g. opening the widget when tapping on it). Looks like it crashed completely. But on iOS, usually when an app crashes, it gets closed completely. This one stays open.

It's only happened on iOS, Android works fine.

skddc:
So only the widget freezes, but not the rest of the app?

galfert:
No, it's the whole app.

skddc:
And you're sure it's not every time, but only sometimes? I.e. the situation is also not the same every time, like e.g. only when the token exists on the server already or something like that?

It would seem like an issue with the in-app browser plugin to me in general.

Does it crash in the simulator, too?

MrPutuLips:
I have tested this on a cordova app and it is working well. I haven't experienced any freezes as mentioned above, but I am not using the remotestorage widget, so I'm not sure if that is the issue.

skddc:
@galfert Ping. (You said on chat that you'd try the simulator.)

skddc:
Bump.

galfert:
Oh, to be honest I've forgotten about this one. I've put it high on my todo list.

galfert:
I tested this with a new Cordova app and this time all went fine. No crashes or freezes.

So I think this is good to go.

src/remotestorage.js
skddc: This is still unanswered.

skddc:
Yay!

Just the documtation issue left now.

src/remotestorage.js
galfert: I added a note that this function is not called by an app directly but rather from a custom storage backend implementation.

galfert:
I added a note to the function documentation. I think this is ready to merge now.

src/remotestorage.js
```diff
@@ -174,6 +174,9 @@ RemoteStorage.prototype = {
   /**
    * Initiate the OAuth authorization flow.
    *
+   * This function is usually not called by an app directly, but rather from a
+   * custom storage backend implementation (e.g. Dropbox or Google Drive).
+   *
```
skddc: Is there actually an unusual case for using this directly? If not, shouldn't we make it private API?

galfert: I don't know. Maybe if you want your app to be tied to one specific storage backend directly you might want use this function.

galfert: I didn't mean just a general RS backend, but one specific server, e.g. a RS server that is specific to the app.

skddc: I don't think I can follow. But I don't we should encourage people to use app-specific servers. It's also still not clear why one would use that function in this case.

Please let's be completely clear about this, and if there's no good case for using it then let's make it private and remove any confusing comments hinting at use cases without actually explaining a single one.

galfert: Ok, I'll mark it as private and adapt the description.

PR #1102, opened 2017-11-24
Fix offline app sync error

galfert:
Sometimes, when offline, instead of the `ProgressEvent` error the request is rejected because of a timeout instead (at least in Firexfox).

This change handles that case and prevents the `Sync` error being emitted, as reported in remotestorage/remotestorage-widget#69.

I added two test cases for both the `ProgressEvent` and the timeout case. I also fixed all the linting errors for the `dropbox-suite.js` test file. For that I had to change the indentation, so the diff shown in here is for the whole file. But that's a different commit, the commit with the fix just contains the actual tests added.

### DIFF
--- a/src/dropbox.js
+++ b/src/dropbox.js
@@ -810,7 +810,11 @@ Dropbox.prototype = {
     self._revCache = new LowerCaseCache('rev');
 
     return fetch().then(undefined, function (error) {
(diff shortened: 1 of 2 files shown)


skddc:
Very cool! Code lgtm and verified the fix with the widget demo.

PR #1103, opened 2017-11-24
Caching (config) improvements

skddc:
* Document caching.checkPath() as public function
* Refactor caching suite, fixes eslint errors


galfert:
👍

PR #1105, opened 2017-11-29
Remove ESLint max-lines limit, update dependencies

skddc:
* Remove max-lines limit
* Update eslint and babel-core
* Fix indentation in CodeClimate config

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -7,7 +7,7 @@ engines:
     enabled: true
     config:
       languages:
-      - javascript
+        - javascript
 ratings:
   paths:
     - src/**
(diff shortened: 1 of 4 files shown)


skddc:
@galfert Could you run this through your local CodeClimate to see if the `max-lines` error is gone (e.g. sync.js)?

galfert:
No, the error was still there. I noticed that it's not labeled with `[eslint]`, but with `[structure]`.
Took a while to find it in the documentation, but I found the name of the check and disabled it in the Codeclimate config.

skddc:
Oh, ok. Thanks!

skddc:
Rebase and merged locally.

PR #1106, opened 2017-11-30
Refactorings

galfert:
Small refactorings to remove some duplication.

### DIFF
--- a/src/googledrive.js
+++ b/src/googledrive.js
@@ -33,6 +33,7 @@ const isFolder = util.isFolder;
 const cleanPath = util.cleanPath;
 const shouldBeTreatedAsBinary = util.shouldBeTreatedAsBinary;
 const readBinaryData = util.readBinaryData;
+const getJSONFromLocalStorage = util.getJSONFromLocalStorage;
 
 let hasLocalStorage;
 
@@ -142,12 +143,7 @@ const GoogleDrive = function (remoteStorage, clientId) {
   hasLocalStorage = util.localStorageAvailable();
 
   if (hasLocalStorage){
-    let settings;
-    try {
-      settings = JSON.parse(localStorage.getItem(SETTINGS_KEY));
-    } catch(e) {
-      // no settings stored
-    }
+    const settings = getJSONFromLocalStorage(SETTINGS_KEY);
     if (settings) {
       this.configure(settings);
     }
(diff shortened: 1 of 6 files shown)


src/googledrive.js
```diff
@@ -93,6 +94,18 @@ function googleDrivePath (path) {
...
+/**
+ * Remove surrounding quotes from a string.
+ *
+ * @param {string} string - string with surrounding quotes
+ * @returns {string} string without surrounding quotes
+ *
+ * @private
+ */
+function removeQuotes (string) {
+  return string.substring(1, string.length-1);
```
skddc: This is not actually removing just surrounding quotes, but the first and last character, no matter what it is. Not sure that's a good idea, at the very least not with this function name.

src/remotestorage.js
```diff
@@ -12,6 +12,7 @@ const log = require('./log');
 const Features = require('./features');
 const globalContext = util.getGlobalContext();
 const eventHandling = require('./eventhandling');
+const getJSONFromLocalStorage = util.getJSONFromLocalStorage;
```
skddc: As we have to import all of `util` right now, growing it more and more seems like it's not a great idea. What do you think about making this its own import?

src/util.js
```diff
@@ -242,6 +242,23 @@ var util = {
...
+   * @param {string} key - localStorage key
+   *
+   * @returns {object} parsed object or undefined
+   */
+  getJSONFromLocalStorage (key) {
+    const context = util.getGlobalContext();
+
+    try {
+      return JSON.parse(context.localStorage.getItem(key));
+    } catch(e) {
```
skddc: Maybe an explicit `null` would be a better return value in this case.

However, both could be confusing, if someone edits their cached data manually and has a JSON error in it. Now that it's a seperate function, we could also actually check if the key exists and then make it throw an exception if it does, and contains data (not an empty string or sth), which is not JSON-formatted.

skddc:
@galfert Did you get a chance to look at my comments?

src/remotestorage.js
galfert: I agree in general, but this file needs to import all of `util` anyway, as it's setting it as `RemoteStorage.util` in [line 773](https://github.com/remotestorage/remotestorage.js/blob/refactorings/src/remotestorage.js#L773).

src/util.js
galfert: This would change the functionality though. I'm not opposed to doing this, but this PR is just supposed to contain some refactorings that don't change any logic.

Should I create a separate issue for this?

skddc:
applause

PR #1107, opened 2017-12-02
Integrate TravisBuddy

bluzi:
TravisBuddy will comment on pull requests in your public repository everytime a test failed in one of them. The comment will include only the part of the build log that applies to your testing framework, so your contirbutors won't have to enter Travis's website and search the long and annoying build log for the reason the tests failed.

Here's an example: https://github.com/bluzi/static-server/pull/1

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,3 +11,12 @@ sudo: false
 script:
     - npm test
     - npm run lint
+notifications:
+  webhooks:
+      urls:
+        - https://www.travisbuddy.com/
+      on_success: never
+      on_failure: always
+      on_start: never
+      on_cancel: never
+      on_error: never


skddc:
We don't really need a hosted service to add information about something that GitHub already adds. If you think that TravisBuddy actually parses Jaribu output in a nice way, then maybe let this pull request fail as a demo?

bluzi:
Hey @skddc!

Unfortunately TravisBuddy support prettifying test logs at the moment, but it's a goal we're working on.
However, TravisBuddy adds information beyond what GitHub supports - While GitHub only indicates whether the checks failed or succeeded, TravisBuddy imminently presents the logs, and makes it easier for contributors to identify what's wrong with their changes. 

Here's an example from one of my repositories, where TravisBuddy's comment made the contributor easily see what breaks the tests, it saves me the effort of commenting myself, because some contributors just won't notice the tests failed or just won't understand why, because the log is too complicated.
https://github.com/bluzi/name-db/pull/469

However, if you feel like it's not necessary here, feel free to close this PR :)

skddc:
Ok, so it doesn't work here, because the output would contain pages upon pages of log before the failure. I'll close the PR then. Thanks.

PR #1110, opened 2017-12-31
Improve error event documentation

skddc:
Also imrove the rendering of tables on large screens. See commmits.


galfert:
seal_of_approval

PR #1111, opened 2017-12-31
Remove obsolete utils

skddc:
These functions are not used anymore it seems. If they're used in a legacy module, the code has to be moved to a new utility module.


galfert:
👍

Issue #1112, opened 2018-01-08
Bower install failed


yassinphilip:
Hi ; [The manual states](https://remotestoragejs.readthedocs.io/en/latest/getting-started/how-to-add.html#adding-rs-js-to-an-app) that `remotestoragejs` is installable via bower, but there seem to be a problem with that :

```
./node_modules/bower/bin/bower install -S remotestoragejs
bower                        ENOTFOUND Package remotestoragejs not found

```

skddc:
Oops. According to https://bower.io/search/ that might be just `rs` instead of `remotestoragejs`. Could you try that?

yassinphilip:
Oooh :)


```
# ./node_modules/bower/bin/bower install -S rs             
bower rs#*                  not-cached https://github.com/remotestorage/remotestorage.js.git#*
bower rs#*                     resolve https://github.com/remotestorage/remotestorage.js.git#*
bower rs#*                    download https://github.com/remotestorage/remotestorage.js/archive/v1.0.2.tar.gz
bower rs#*                     extract archive.tar.gz
bower rs#*                    resolved https://github.com/remotestorage/remotestorage.js.git#1.0.2
bower remotestorage#^1.0.2     install remotestorage#1.0.2

remotestorage#1.0.2 bower_components/remotestorage

```

I guess someone should update the manual then :package: 
Thanks a lot, @skddc

skddc:
(Btw, the manual has an "edit on Github" link in the upper right corner, btw. In case you want to get your first contribution in. 😉)

yassinphilip:
> In case you want to get your first contribution in.

/bin/done :)

What about the widget then? Is there a bower install path?

skddc:
> What about the widget then? Is there a bower install path?

Good point. I think nobody registered a bower package for that yet.

PR #1113, opened 2018-01-08
Update how-to-add.rst

yassinphilip:
Updated the install command according to https://github.com/remotestorage/remotestorage.js/issues/1112


skddc:
Cool, thanks!

PR #1118, opened 2018-03-14
Don't fail when sync instance is not defined

### LINKED ISSUE #1117: Cannot read property 'stopped' of undefined
I'm getting this error in an integration test for an app that uses remoteStorage.js.

In that test, the app connects to remoteStorage and then pretty much immediately disconnects again. The error doesn't occur on every test run, but happens more often than not.

galfert:
Fixes #1117

### DIFF
--- a/src/remotestorage.js
+++ b/src/remotestorage.js
@@ -701,13 +701,13 @@ RemoteStorage.prototype = {
    * @private
    */
   syncCycle: function () {
-    if (this.sync.stopped) {
+    if (!this.sync || this.sync.stopped) {
       return;
     }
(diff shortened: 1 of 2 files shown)


src/remotestorage.js
```diff
@@ -701,13 +701,13 @@ RemoteStorage.prototype = {
    * @private
    */
   syncCycle: function () {
-    if (this.sync.stopped) {
+    if (!this.sync || this.sync.stopped) {
```
skddc: Why not just use the same as below, which I think is a bit more standard?

PR #1119, opened 2018-03-14
Don't close IndexedDB DB while still writing to it

### LINKED ISSUE #939: Sync error after disconnect

galfert:
Waits for the writes to finish first before closing the DB.

Fixes #939

### DIFF
--- a/src/indexeddb.js
+++ b/src/indexeddb.js
@@ -274,7 +274,11 @@ IndexedDB.prototype = {
   },
 
   closeDB: function () {
-    this.db.close();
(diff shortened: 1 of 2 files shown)


skddc:
thumbs-up-machine

silverbucket:
+1, and +1 that gif too

PR #1121, opened 2018-03-24
Fix remote.storageApi not being set correctly

### LINKED ISSUE #1120: storageApi is not set after connecting to 5apps.com
While investigating a case of missing 'conflict' events, I found that in my case no If-Match headers got used. After adding some additional debug output, I noticed that no storageApi is set and thus supportsRevs is false when [calling WireClient.put](https://github.com/remotestorage/remotestorage.js/blob/master/src/wireclient.js#L460).

This happens for me both with Firefox and Chrome. Here are the request headers when using [favorite drinks](https://myfavoritedrinks.remotestorage.io):

```
PUT /karlb/myfavoritedrinks/test HTTP/1.1
Host: storage.5apps.com
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.11; rv:59.0) Gecko/20100101 Firefox/59.0
Accept: */*
Accept-Language: en-US,en;q=0.7,de;q=0.3
Accept-Encoding: gzip, deflate, br
Referer: https://myfavoritedrinks.remotestorage.io/
Content-Type: application/json; charset=UTF-8
Authorization: Bearer [REMOVED]
Content-Length: 88
Origin: https://myfavoritedrinks.remotestorage.io
Connection: keep-alive
```

skddc:
Some places that should be using storageApi are setting storageType instead. But the latter is a different property, albeit related in functionality.

This fixes a serious problem with versioning, which is that due to storageApi not being set in the remote/wireclient, no `If-Match` headers are sent in PUT requests.

fixes #1120

### DIFF
--- a/src/discover.js
+++ b/src/discover.js
@@ -21,7 +21,7 @@ var cachedInfo = {};
  *
  * @returns {Promise} A promise for an object with the following properties.
  *          href - Storage base URL,
- *          storageType - Storage type,
+ *          storageApi - RS protocol version,
  *          authUrl - OAuth URL,
  *          properties - Webfinger link properties
  **/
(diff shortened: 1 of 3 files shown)


galfert:
I was able to reproduce the original problem and can confirm that this PR fixes the bug.
Changes look good to me, too.

👍

PR #1122, opened 2018-03-29
Update dependencies, build config

skddc:
The Webpack upgrade from version 3 to 4 requires some changes in config. I tested both development and production sourcemaps in Firefox.


galfert:
Works for me and looks all good. Sourcemaps work in Chrome as well. 

👍

PR #1123, opened 2018-04-05
Fix return value for storeFile doc

skddc:
It's not actually an object, but the ETag as a string, as obvious from the code directly in this function.


galfert:
Looks good 👍

PR #1127, opened 2018-05-23
Set output.globalObject option in Webpack config

### LINKED ISSUE #1126: No longer works in web worker since version 1.0.3
Hey folks 👋 

I was using version `1.0.2` in my project and decided to upgrade to `1.0.3`.
But now I'm getting the following error:

```js
Uncaught ReferenceError: window is not defined
```

The issue only pops up in web workers.

I'm using remoteStorage through browserify and I'm requiring it as follows:

```js
self.RemoteStorage = require("remotestoragejs/release/remotestorage");
```

Any thoughts?
Thanks!

galfert:
Fixes #1126

This fixes a reference error when using the build in a WebWorker.

This is a workaround taken from https://github.com/webpack/webpack/issues/6642#issuecomment-371087342 and should be replaced by `target: "universal"` when https://github.com/webpack/webpack/issues/6525 has been implemented.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -22,7 +22,8 @@ module.exports = {
     // global export name if needed
     library: 'RemoteStorage',
     libraryTarget: 'umd',
-    umdNamedDefine: true
+    umdNamedDefine: true,
+    globalObject: 'this'
   },
   // the only external dependecy is xmlhttprequest because it is
   // different in browser and in node env so user has to manage with that


skddc:
How can I actually review this?

galfert:
I guess you could check that the build is using `this` instead of `window` again after this has been applied. See https://github.com/remotestorage/remotestorage.js/issues/1126#issuecomment-391050035.

PR #1129, opened 2018-06-01
Make Dropbox correctly fetch folders

### LINKED ISSUE #1128: Dropbox backend does not pick up remote changes
Reading a file with a path of say `/foo` from the Dropbox backend correctly fetches and caches the folder listing for `/` and the file `/foo` itself.

However, when I then change the `foo` file in the Dropbox, I don't get any `change` events in my test app.

JulianKniephoff:
This fixes #1128.

I have spent quite some time to figure this out and I still can't quite put all of it to words, but the problem seems to be the following:

When rs.js drops with the Dropbox backend, it first gets a recursive listing of everything under the storage prefix (`/remotestorage` for me). This is used to fill/update the revision cache.

Then the usual task collection begins and (in the example from the issue) determines that `/` should be refreshed, let's say. This is delegated to `Dropbox#get`, which is given the current revision of `/` (which is just `"rev"`, the default value of the revision cache, because `/` is a folder). The "preflight" listing mentioned above again wrote `rev` to the revision cache for the path `/` and so the `get` function immediately determines there is nothing to do and returns a status of 304 (nothing changed).

This might not be true, though. The backend **has** to check the folder listing for `/` again, to see whether any of its items changed. `_getFolder` does this correctly, so I just reordered the "early 304" and the delegation to `_getFolder`.

I am not sure whether this is the "morally correct" solution, so any input is very much appreciated. :smile:

### DIFF
--- a/src/dropbox.js
+++ b/src/dropbox.js
@@ -375,6 +375,11 @@ Dropbox.prototype = {
     var url = 'https://content.dropboxapi.com/2/files/download';
     var self = this;
 
+    //use _getFolder for folders
+    if (path.substr(-1) === '/') {
+      return this._getFolder(path, options);
+    }
+
     var savedRev = this._revCache.get(path);
     if (savedRev === null) {
       // file was deleted server side
@@ -384,11 +389,6 @@ Dropbox.prototype = {
        savedRev && (savedRev === options.ifNoneMatch)) {
       // nothing changed.
       return Promise.resolve({statusCode: 304});
-    }
-
-    //use _getFolder for folders
-    if (path.substr(-1) === '/') {
-      return this._getFolder(path, options);
     }
 
     var params = {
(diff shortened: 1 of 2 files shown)


skddc:
Hey, thanks for the pull request and welcome to the community!

I'm trying to wrap my head around why it would return 304s from cache before even checking the server there, but I couldn't come up with an explanation. Unfortunately, the Git history also doesn't have an answer. Those lines have existed for quite a long time, and the commit message back then was "syncing bugfixes". :clown_face: 

@Ragnis @galfert Maybe one of you has an idea why it's doing that there?

galfert:
I also don't know why it is the way it is right now. But the PR indeed fixes the problem and the solution looks good to me as well.

Thank you 👍

galfert:
FYI: I added a changelog entry for the fix.

JulianKniephoff:
Cool, thanks for the quick merge! :smile:

Issue #1052, opened 2017-10-27
Problems with getItemURL on Dropbox


galfert:
To get the public URL for a shared file in Dropbox, the client needs to make a request to Dropbox to create/get that URL. Since that is asynchronous, the Dropbox client keeps a local cache (`_itemRefs`) of those URLs which is then used by `getItemURL`.

But the cache is only filled when either GETing or PUTing a file. So when a directory is cached, the URLs are only available after the first sync cycle. But when caching is disabled, `getItemURL` always returns and empty string.

galfert:
Also, for private files, `getItemURL` just doesn't return anything.

PR #1130, opened 2018-06-02
Implement getItemURL for Dropbox

Ragnis:
#1052

### DIFF
--- a/doc/getting-started/dropbox-and-google-drive.rst
+++ b/doc/getting-started/dropbox-and-google-drive.rst
@@ -49,7 +49,7 @@ Known issues
 * Listing and deleting folders with more than 10000 files will cause problems
 * Content-Type is not fully supported due to limitations of the Dropbox API
 * Dropbox preserves cases but is not case-sensitive
-* ``getItemURL`` is not implemented yet (see issue :issue:`1052`)
+* ``getItemURL`` works only for files which exist and are public
 
 Google Drive
 ------------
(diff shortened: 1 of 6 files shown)


skddc:
@Ragnis Is this actually still work in progress? Looks to me like this is basically done.

skddc:
@Ragnis Would be cool if you could update us on the status here. See my question from November.

Ragnis:
I guess it is done. I re-based the code to latest master and also fixed the tests.

raucao (changes requested):
I just played around with this branch in Sharesome and found a logical bug with the implementation:

The way the function works on a normal BaseClient is that the path argument is relative to the client scope. So when used in a module, the `publicClient` would have a base of `mymodulename/public/`. Hence the `makePath()` call here: https://github.com/remotestorage/remotestorage.js/blob/master/src/baseclient.js#L299

However, this implementation doesn't add the base path, and so `_itemRefs` aren't matched correctly when calling `getItemURL()` with a relative path (in the shares module that would just be a filename e.g.). This results in `undefined` being returned instead of the stored public URL from `_itemRefs`.

---

Another thing, but I'm not 100% sure how to solve it, is that `storeFile()` actually returns a URL on a normal BaseClient, but we only overwrite `put` in the Dropbox class, and returning a simple URL from that broke it for me when I tried to add it.

raucao:
Closing in favor of #1389

PR #1132, opened 2018-10-28
wireclient uses fetch() if available, instead of XHR

DougReeder:
The primary need for this is running inside a Service Worker, where XHR is unavailable.  It's also of use in Node, when a fetch() implementation is preferred over an XHR. 

Secondarily, it allows timed-out requests to be aborted at the network level, so they don't continue to consume resources.


src/wireclient.js
```diff
@@ -490,8 +490,81 @@ WireClient.prototype = {
...
+        return response.arrayBuffer();
+      case 'blob':
+        return response.blob();
+      case 'json':
+        return response.json();
+      case '':
+      case 'text':
+        return response.text();
+      default:   // document
+        throw new Error("responseType 'document' is not currently supported using fetch");
```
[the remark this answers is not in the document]
DougReeder: This is controlled by the calling function, which sets options.responseType.  It is independent of the Content-type returned by the server.

So far as I can tell, all current code always sets options.responseType to 'arraybuffer' and converts as needed, outside WireClient.request.  However, WireClient.request is a public interface, and 'blob', 'json' and 'text' are plausible values (and easily implemented).

'document' is used by XHR to produce an HTML Document or XML XMLDocument.  It is possible to write code to produce those from the fetch response, but I can't forsee any use for that.  That would be something UI code would use.

.gitignore
```diff
@@ -9,3 +9,5 @@ server-state
 doc/code
 doc/_build
 /remotestorage*.js
+.idea/
+remotestorage.js.iml
```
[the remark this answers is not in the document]
DougReeder: My second commit reverts this.

src/util.js
```diff
@@ -35,10 +35,10 @@ var util = {
     }
   },
 
-  globalContext: (typeof(window) !== 'undefined' ? window : global),
+  globalContext: (typeof(window) !== 'undefined' ? window : (typeof self === 'object' ? self : global)),
 
   getGlobalContext () {
-    return (typeof(window) !== 'undefined' ? window : global);
+    return (typeof(window) !== 'undefined' ? window : (typeof self === 'object' ? self : global));
```
galfert: What is this new change for?

DougReeder: This sets the global context, when run in a Service Worker, where the global variable is called 'self'.

DougReeder:
Are there any other concerns?

skddc:
I'll try to find a cycle for reviewing this tomorrow. We need to release some fixes asap anyway.

src/wireclient.js
```diff
@@ -490,8 +490,81 @@ WireClient.prototype = {
...
+    }
+  }).then(function (processedBody) {
+    syntheticXhr.response = processedBody;
+    return syntheticXhr;
+  });
+
+  var timeoutPromise = new Promise(function (resolve, reject) {
+    setTimeout(function () {
+      log('[WireClient] request timed out', url);
```
skddc: As this promise is raced, but nothing is cancelled, this code will still be run. I'm not sure about potential side effects (for example I had rs-backup not exit the program for me after syncing all items with this branch, but not sure if related). However, the line is always logged in the debug logs, which is definitely not a good idea:

screenshot from 2018-11-17 11-16-24

DougReeder: Good catch!  That log() line should be removed.

Unfortunately, I think I rebased at the wrong time before pushing the commit with this. I'll make a new branch and PR with just my changes.

DougReeder:
closed in favor of #1140

PR #1134, opened 2018-10-31
Refactor synced GET function

### LINKED ISSUE #1133: Argument 'maxAge' must be false or a number
Whenever I minify the following example, remotestorage.js throws an error:

```js
import RemoteStorage from 'remotestoragejs'
const remoteStorage = new RemoteStorage({logging: true})
remoteStorage.access.claim('foo', 'rw')

const client = remoteStorage.scope('/foo/')
client.getListing('', false).then(listing => console.log(listing))
// error: "Argument 'maxAge' must be false or a number"
```

Note, this only happens after building the code. If you don't include remotestorage in the bundle but use a global via cdn or so, there is no error. I am not connecting to any remote storage, e.g. only local.

Could it be how the code is minified? You can test by putting above example inside a `create-react-app` (webpack) or `npx bili example.js --format umd` (rollup).

May be related to something here https://github.com/remotestorage/remotestorage.js/blob/master/src/syncedgetputdelete.js#L25

skddc:
Simplify the code. Aside from being more readable, this fixes a problem with the UglifyJS compressor destroying the code by restructuring it wrong.

closes #1133

### DIFF
--- a/src/syncedgetputdelete.js
+++ b/src/syncedgetputdelete.js
@@ -5,29 +5,27 @@ function shareFirst(path) {
            path.match(/^\/public\/.*[^\/]$/) );
 }
 
-function maxAgeInvalid(maxAge) {
-  return maxAge !== false && typeof(maxAge) !== 'number';
+function defaultMaxAge(context) {
+  if ((typeof context.remote === 'object') &&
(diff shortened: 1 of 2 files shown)


skddc:
Closing, because the first commit doesn't belong here.

PR #1139, opened 2018-11-17
Update dependencies

skddc:
* Update eslint, fix one new linting issue
* Update babel loader and Webpack config to new packages/names

Also removes the CodeClimate badge from the README

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,0 +1,6 @@
+# Browsers that we support
+
+cover 99.5%
+ie >= 11
+maintained node versions
+not dead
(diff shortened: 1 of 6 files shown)


galfert:
For that I used `cover 99.5% in DE` as the query because it made sense for that particular app, but I don't think that's the right approach for `rs.js`.

I would suggest something like

```js
{
  "targets": {
    "browsers": ["last 3 versions", "ie >= 11"],
    "node": "maintained node versions"
  }
}
```

The [documentation](https://babeljs.io/docs/en/next/babel-preset-env.html#debug) mentions a debug mode that logs the plugins used with the given config when compiling.

skddc:
Thanks, I see. The important part of the question was about the default however. I don't know if your custom config covers more or less browsers.

"Last 3 versions" is not a lot with evergreen browsers. It wouldn't even cover the last 2 years in full. I think that's a bit harsh, considering not all end users have auto-updates enabled and we should aim to keep compatibility roughly at where it was (which I think is pretty much everything with CORS and IndexedDB).

skddc:
I don't see why that would have any benefits, but the drawbacks are clear: browserslist automatically includes anything that is both used in the code and not supported by any of the browsers configured. No need to forget adding configs while coding in ES6.

galfert:
From https://github.com/browserslist/browserslist#queries:

> If the above methods did not produce a valid result Browserslist will use defaults: `> 0.5%, last 2 versions, Firefox ESR, not dead`.

skddc:
Ok, so I added a `.browserslistsrc`, which is the recommended way of doing this, according to https://new.babeljs.io/docs/en/next/babel-preset-env.html

I played around with it a little bit, and short of using a hardcore setting like "cover 10%" the size never changed from 144KB. And the 10% setting only brought it down to 140KB. So I think this is a reasonable config:

```
cover 99.5%
ie >= 11
maintained node versions
not dead
```

galfert:
Looking good 👍

PR #1140, opened 2018-11-18
wireclient uses fetch() if available, instead of XHR

DougReeder:
Consolidated requested updates with initial request and rebased to incorporate latest changes from master.


skddc:
@DougReeder I added you as collaborator to the remoteStorage organization, so you can also create and push to branches of the shared repositories from now on.

PR #1147, opened 2018-12-17
Refactoring sync.js

skddc:
Trying to clean up sync in preparation of adding cached item metadata and improving sync behavior. Work in progress.

### DIFF
--- a/src/sync.js
+++ b/src/sync.js
@@ -11,7 +11,7 @@ const equal = util.equal;
 const deepClone = util.deepClone;
 const pathsFromRoot = util.pathsFromRoot;
 
-function taskFor(action, path, promise) {
+function taskFor (action, path, promise) {
   return {
     action:  action,
     path:    path,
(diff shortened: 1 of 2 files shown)


test/unit/sync-suite.js
```diff
@@ -339,8 +330,8 @@ define(['require', 'test/helpers/mocks'], function(require, mocks) {
             });
           };
           env.rs.sync.numThreads = 5;
-          env.rs.sync.remote.connected = true;
-          env.rs.sync.remote.online = true;
+          env.rs.sync.rs.remote.connected = true;
+          env.rs.sync.rs.remote.online = true;
```
galfert: Why the loop?

```suggestion
          env.rs.remote.online = true;
```

src/sync.js
```diff
@@ -118,13 +110,13 @@ Sync.prototype = {
 
   queueGetRequest: function (path) {
     return new Promise((resolve, reject) => {
-      if (!this.remote.connected) {
+      if (!this.rs.remote.connected) {
```
galfert: Instead of all the changes to `this.rs.remote` and `this.rs.local` everywhere, I would rather set a local reference (`this.remote = this.rs.remote`) and keep accessing it via `this.remote`.

skddc: That's exactly how it worked before (except for unnecessary constructor arguments). If you read my commit description, then you'll find that this is intentional in order to make it clear that this is a remoteStorage instance's remote, and also make it easier to rip that out in the future, because why should there be `remoteStorage.sync.remoteStorage` in the first place.

skddc:
As this is actually green, and already quite a few lines of change, and also because this fixes [an error found in another, rather important, open PR](https://github.com/remotestorage/remotestorage.js/pull/1161#discussion_r256622416), I'm removing the WIP and propose to merge this as is.

galfert:
What about this suggestion? https://github.com/remotestorage/remotestorage.js/pull/1147#discussion_r242466440

Other than that I'm fine with merging.

Issue #1151, opened 2019-01-05
Fetch API breaks Dropbox and GoogleDrive support


iLiviu:
The problem is that if no `responseType` property is present in `options ` parameter to `WireClient._fetchRequest`, the function defaults to `document` and throws an error. If `responseType `is `undefined`, the function should treat the response as `text`. Also, if the response is treated as `text` , the `syntheticXhr.responseText` property should be set to `processedBody` in order to not break XHR compatibility.

skddc:
Merged and released. Thanks for reporting and fixing, @iLiviu!

PR #1152, opened 2019-01-07
fix default case for responseType in WireClient._fetchRequest #1151

iLiviu:
If no responseType property is present in options parameter to WireClient._fetchRequest,
the function defaults to document and throws an error. If responseType is undefined,
the function should treat the response as text. Also, if the response is treated as text,
the syntheticXhr.responseText property should be set to processedBody in order to not
break XHR compatibility.

### DIFF
--- a/src/wireclient.js
+++ b/src/wireclient.js
@@ -536,6 +536,7 @@ WireClient._fetchRequest = function (method, url, options) {
         return response.blob();
       case 'json':
         return response.json();
+      case undefined:
       case '':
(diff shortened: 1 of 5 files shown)


iLiviu:
i will try and refactor the dropbox-suite and googledrive-suite and include tests for fetch when i get some time

skddc:
I can confirm that this fixes both Dropbox and GDrive when compared to `1.2.0`. Code lgtm.

PR #1155, opened 2019-01-20
fix GoogleDrive and Dropbox getFile returning corrupted binary data

### LINKED ISSUE #1116: Google Drive and Dropbox images getFile corrupted
Been trying to load images added to Drive or Dropbox via `storeFile`, however `getFile` is loading corrupt data.

```js
remoteStorage.scope('/images/')
.getFile('example.png').then(file => {
  // Neither of these work

  // try to create object url
  var blob = new Blob([file.data], { type: file.contentType });
  let bleh = URL.createObjectURL(blob);
  this.rsData.push(bleh);
  URL.revokeObjectURL(bleh);

  // try to create Data URL
  var reader = new FileReader();
  reader.onloadend = (e) => {
    this.rsData.push(e.target.result);
  };
  reader.readAsDataURL(blob);
});
```

Tried a ton of different combinations, the images appear correctly when viewing in google drive and dropbox UI, so I think its a problem with the getFile function.

```js
// this is what the image should start with as a data url
data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAArwAAAK8CAMAA...

// this is what keeps coming thru, the corrupt versions all seem to start with "77+"
data:image/png;base64,77+9UE5HDQoaCgAAAA1JSERSAAAC77+9AAAC77...
data:image/jpeg;base64,77+977+977+977+9ABBKRklGAAEBAQEsASwAAO...
```

Any help would be much appreciated. Let me know if you have any follow up questions. Would love to get this figured out.

using `"remotestoragejs": "^1.0.2"`

iLiviu:
Fixes #1116 
Dropbox and Googledrive backends first read the response as text and then tried to convert it back to raw value, however not all bytes can be encoded as text, and are lost during the conversion to text and back. Now, they request the response as `ArrayBuffer` and encode it to text and if they detect binary data, they return the unprocessed `ArrayBuffer`.
Also made a fix to `shouldBeTreatedAsBinary` function because it was treating white space characters (horizontal tab, vertical tab, CR, LF) as binary specific characters and added the Unicode replacement character (\uFFFD) as binary specific because the decoder inserts this character when it detects an invalid code sequence for the text encoding provided.

### DIFF
--- a/src/dropbox.js
+++ b/src/dropbox.js
@@ -45,8 +45,8 @@ const PATH_PREFIX = '/remotestorage';
 const isFolder = util.isFolder;
 const cleanPath = util.cleanPath;
 const shouldBeTreatedAsBinary = util.shouldBeTreatedAsBinary;
-const readBinaryData = util.readBinaryData;
 const getJSONFromLocalStorage = util.getJSONFromLocalStorage;
+const getTextFromArrayBuffer = util.getTextFromArrayBuffer;
 
 /**
  * Map a local path to a path in Dropbox.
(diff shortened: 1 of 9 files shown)


skddc:
+1

galfert:
Added a changelog entry for this in 53e72717.

PR #1158, opened 2019-02-04
Fix detecting changes in subfolders for Dropbox

### LINKED ISSUE #1154: Dropbox backend does not pick up remote changes for subfolders
The fix #1129 for issue #1128 only fixes picking up remote changes for root folder. If the root folder has subfolders, and the content of a subfolder changes remotely, the change is not picked up by the backend. There was some strange code i couldn't understand why it was there but when i investigated the history of dropbox.js i discovered that initially it was using Dropbox V1 API. The problems appeared once the code was switched to V2 API which no longer offers revision ids for folders. 

As a solution, i propose hashing the contents of the folder locally in the client and use that as a revision id. I can work on this.

iLiviu:
Fixes #1154 

I modified the revision cache for files that Dropbox was using until now, and made it generate folder revisions by hashing the revisions of the files in it. The change of revision propagates to parent folders, and forces their revision to update so changes in subfolders can now be detected.

### DIFF
--- a/src/dropbox.js
+++ b/src/dropbox.js
@@ -3,7 +3,7 @@ var BaseClient = require('./baseclient');
 var WireClient = require('./wireclient');
 var util = require('./util');
 var eventHandling = require('./eventhandling');
-var Sync = require('./sync');
+var RevisionCache = require('./revisioncache');
 
 /**
  * WORK IN PROGRESS, NOT RECOMMENDED FOR PRODUCTION USE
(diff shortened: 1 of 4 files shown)


skddc:
I reviewed and tested this with an app, and it works like a charm. Good job!

@iLiviu I have a commit that cleans up formatting and updates syntax a bit in the new class, but I can't push it to your repo, of course. I just invited you to become a member of this repository, so you can push branches and create pull requests directly over here. This way, we can collaborate on pull requests. Would you mind pushing your branch here, and re-opening another PR, so I can add my commit?

iLiviu:
Thank you for the invitation. Looking forward to future collaboration.

PR #1160, opened 2019-02-05
do not return common revision if local node was marked for deletion

### LINKED ISSUE #1150: BaseClient.remove not working correctly while offline
When connected to a backend and working offline, if a node is removed by calling `BaseClient.remove`, calling `getObject`/`getFile` afterwards will still return the last common revision.

iLiviu:
Fixes #1150 

If BaseClient.getObject/getFile were called after a file has been removed, but before a successful sync, they were still returning the common revision of the file, if present. This happened because the cachingLayer never checked if the local version exists but was deleted.

This issue was especially problematic when working offline and no sync was done to flush local version.

### DIFF
--- a/src/cachinglayer.js
+++ b/src/cachinglayer.js
@@ -32,8 +32,13 @@ function getLatest(node) {
       return node.common;
     }
   } else {
-    if (node.local && node.local.body && node.local.contentType) {
-      return node.local;
+    if (node.local) {
+      if (node.local.body && node.local.contentType) {
+        return node.local;
+      }
+      if (node.local.body === false) {
+        return;
+      }
     }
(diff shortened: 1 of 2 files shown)


iLiviu:
This can only be reproduced successfully on an application that works offline, because as soon as a successful sync is completed, the local changes are pushed remotely and file removed. When working offline, problem persists even if the page is reloaded.

skddc:
I tried it with the latest version of Webmarks, which does work fully offline. But I just realized, the reason for it not loading the bookmark on the edit screen anymore after deletion, is that it keeps the bookmarks collection in memory locally, and that's where it loads the object from.

Issue #1168, opened 2019-10-15
IndexedDB issues on iOS 12 & 13


galfert:
In iOS 12 and 13 there are issue with IndexedDB. The connection gets lost when the Browser has been suspended to the background for a while. After resuming, the IndexedDB can't be accessed for reading or writing. Trying to close/reconnect the DB doesn't work either. Only after a full page reload the DB can be opened again.

This also happens when using native app wrappers with integrated, Webkit-based Webview, like Cordova.

Detailed bug reports can be found at https://bugs.webkit.org/show_bug.cgi?id=197050#c22 and https://bugs.webkit.org/show_bug.cgi?id=202705.

I'd propose to disable IndexedDB usage for the affected versions until Apple has fixed the underlying issue.

skddc:
> I'd propose to disable IndexedDB usage for the affected versions until Apple has fixed the underlying issue.

+1

DougReeder:
When IDB is disabled does it fall back to another storage mechanism?

 If not, it might make sense to tell the user to restart the app or device.

skddc:
It falls back to localStorage.

However, this change would prevent people from accessing unsynced data, so I think it needs a major version change, as well as proper announcement/explanation for app developers.

rosano:
Is there a way to detect this suspended state? Or is there code in the rs library that runs when the saving 'fails' because if this? I would like hotfix/patch it for myself to reload the page immediately. I lost data several times because of this issue and didn't know what the problem was.

skddc:
What do you mean by "suspended state"? The iOS bugs are not in the library, or detectable by the library, unfortunately. Apple just failed their users spectacularly (and not for the first time with IndexedDB on iOS I should add).

We should probably still release the proposed "fix", but so far nobody proposed how to solve the loss of access to unsynced data if we simply merge this PR. @rosano Maybe you have an idea for how to handle this?

/cc @galfert

rosano:
It's undetectable? I understand the connection can't be restarted and it isn't possible to read/write, but is there a way to at least (perhaps periodically) inspect whether the connection is active/functional? [BTW I have no idea how IndexedDB works!]

Also, I might be misunderstanding but if the PR was merged wouldn't that create a state where existing data is stored in IndexedDB would be ignored and iOS 12/13 users would suddenly be presented with a blank slate (reading from another store)? I imagine one would have to migrate the data to localStorage before switching.

rosano:
[...]

PR #1169, opened 2019-10-15
Don't use IndexedDB on iOS 12 and 13

galfert:
Refs #1168

The regex for the version detection is taken from the [Bowser project](https://github.com/lancedikson/bowser).

### DIFF
--- a/src/indexeddb.js
+++ b/src/indexeddb.js
@@ -246,7 +246,7 @@ IndexedDB.prototype = {
           // hacky!
           this.db = other;
         }
-        if (typeof callback === 'function') { callback(self); }
+        if (typeof callback === 'function') { callback(this); }
       });
     });
   },
--- a/src/indexeddb.js
+++ b/src/indexeddb.js
@@ -403,11 +403,24 @@ IndexedDB._rs_supported = function () {
 
     // Detect browsers with known IndexedDb issues (e.g. Android pre-4.4)
     var poorIndexedDbSupport = false;
-    if (typeof navigator !== 'undefined' &&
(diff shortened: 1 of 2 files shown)


skddc:
Could you explain how this behaves with existing local data?

Especially when having edited something locally, but not being synced with a remote, before opening a new app version that switches from IndexedDB to localStorage. Would the new data and changes survive?

galfert:
If saving the data to IndexedDB actually worked and was not prevented by the bug, I think the data would still be in IndexedDB but not be accessible by the app.

skddc:
If the data isn't accessible by the app anymore, then this is a breaking change and needs to be documented properly, no?

skddc:
FYI: I just published a maintenance release that updates all dependencies and fixes all security alerts, before we potentially publish a major version with this breaking change.

skddc:
@galfert What's the current state on iOS? Is this still a necessary fix, for which we need to release the new major version?

galfert:
It's still not fixed on iOS. The bugs (https://bugs.webkit.org/show_bug.cgi?id=197050#c22 and https://bugs.webkit.org/show_bug.cgi?id=202705) are still open.

skddc:
OK, so how do we proceed then? I think we should not merge this without properly documenting the problem for users who may lose access to their data due to this change. And according to SemVer, this should be a major release also.

skddc:
Just FYI: we'll have to tag a major release soon anyway. So this PR could be part of it.

raucao:
Closing, in favor of @galfert's latest idea, which is to do a quick write check when coming back to the foreground and offer to reload in case it fails.

PR #1175, opened 2020-04-04
Convert source code to TypeScript

skddc:
I think this will make the source code easier to understand and reason about, especially in the very large classes/modules that pass data between many functions. It also removes the potential for type-related bugs, of course. Last but not least, with the popularity of TypeScript, people have already asked for TS definitions for this library, for importing it in their TS codebases.

As of now, I have only converted the Sync class, because that was the one where I was trying to debug and change something at the time. I haven't even started adding types really, but already fixed a bunch of other issues that the TS compiler complained about.

@johannesjo has [offered to help](https://community.remotestorage.io/t/state-of-the-project/614/5?u=raucao) with this, and proposed a mode of collaboration:

> We probably should take good care not to work on it at the same time, as there are so many potential conflicts. Maybe we can do it like this: Once we start working we write a comment on the Draft PR and once we’re done (no unpushed local changes) we write another one?

I like the idea of writing comments with human-readable updates when handing back the code. But I think for managing who's currently working on it, we can just use the assignee feature. Whoever starts on something just self-assigns the PR, and unassigns when they stop pushing changes.

@johannesjo, I just invited you as a collaborator to the RS org, so you can push directly to this branch and also self-assign PRs and issues in any of the org repos.

## To Do

* [x] Convert all source files to TypeScript files (with no compilation/ESLint errors, but warnings allowed)
* [x] Lint TypeScript (instead of JS)
* [x] Update release/versioning process/scripts to include TS compilation
* [ ] Implement process for publishing type definitions
* [ ] Fix/update node.js usage/integration (`index.js`)
* [ ] Ensure the documentation setup still works with the new source files (Sphinx + JSDoc)


skddc:
I added some more types and re-added esformatter. The idea is that we don't commit the JS files for now, until we can get the formatting similar enough to what it was before, in order to preserve `git-blame`-able history.

(I actually do have 2 failing tests for the new SyncError definition locally, but Travis is still testing the old JS code.)

johannesjo:
@skddc maybe it makes sense to get rid of the JavaScript files completely in favor of a single one? Otherwise we might run into the risk of people submitting PRs with changes to them instead of the typescript files.

Or are they needed for backwards compatibility?

skddc:
I would really like to preserve the history. It's helping a lot with this codebase, especially because most of the Sync code e.g. wasn't written by an active maintainer. It's also currently completely tied in with the test suite, but maybe we can change that.

johannesjo:
Makes sense! One step after another :)

johannesjo:
I tried migrating the access.js to typescript there I came across this line:
https://github.com/remotestorage/remotestorage.js/blob/61cbbb25a99e1ce995ddf0fac8571683c32e9b8b/src/access.js#L53

I wonder where is set coming from? Will this be assigned from the outside? Or is the implementation missing?

johannesjo:
I also just recognized that I am not allowed to push to this branch. Should I just create a PR to to it?

skddc:
> I also just recognized that I am not allowed to push to this branch. Should I just create a PR to to it?

No, you should definitely have push access. Let me check again.

skddc:
> I also just recognized that I am not allowed to push to this branch. Should I just create a PR to to it?

No, you should definitely have push access. Let me check again.

skddc:
> I wonder where is set coming from? Will this be assigned from the outside? Or is the implementation missing?

Wow, did you just find a glaring bug in pretty much the only function that's not tested there?

skddc:
I looked at this, and I think the reason the test is missing and there's a bug in the function is that [it is neither a public API](https://remotestoragejs.readthedocs.io/en/latest/js-api/access.html), nor used anywhere else in the code AFAICS.

I also don't see a use case in which we would ever remove a scope like that.

johannesjo:
In this case I just comment it out for now. Maybe we can deal with this in a separate issue.

Not sure if I am doing something wrong regarding the access. I just cloned the repo and tried pushing to this branch directly with my standard credentials. Not sure if that's the right way. I am new to how organizations work on github.

skddc:
Could you post the origin URL that you added for the remote? And the command you use for pushing?

johannesjo:
```
[remote "origin"]
	url = https://github.com/remotestorage/remotestorage.js.git
	fetch = +refs/heads/*:refs/remotes/origin/*
[branch "master"]
	remote = origin
	merge = refs/heads/master
[branch "feature/typescript_conversion"]
	remote = origin
	merge = refs/heads/feature/typescript_conversion

```

skddc:
Oh wait, could it be due to the draft status?

skddc:
> Git then asks me for my credentials, but entering leads to authentication failure.

Authentication failure is between you and GitHub I believe. Shouldn't have anything to do with this repo.

johannesjo:
I should add that the authentication failure is only for this repo. Otherwise it works fine (via public key). Just tried pushing again, no change.

skddc:
> I should add that the authentication failure is only for this repo

Ah, got it. Thanks.

> Just tried pushing again, no change.

Damn. You're in two different teams that both have push access. Looks like it's really a bug in GitHub then. Perhaps some overloaded worker queues or something. :/

You could do a PR against this branch for now, and if it still doesn't work tomorrow, I can contact GitHub's support.

johannesjo:
Now it works. Apparently I had to change `https://github.com/remotestorage/remotestorage.js.git` to `git@github.com:remotestorage/remotestorage.js.git`. No clue why this is the case, but now it works :)

skddc:
I just ran the latest changes locally and the tests are failing due to switching from CommonJS modules to ES modules. The tests are using RequireJS, and also the rest of the lib is using requires, so I think we have to keep using them for now (as sync.ts does, too), and then switch them all at once. It'll be easier to switch them all at once, when we actually have TS files for everything.

I made TS compile on Travis CI before the runs, so that build errors show up in this PR now.

skddc:
Oh, and by the way, you can test single files by using `jaribu` directly, instead of `npm test`, like so:

    jaribu test/unit/localstorage-suite.js

galfert:
> I tried migrating the access.js to typescript there I came across this line:
> https://github.com/remotestorage/remotestorage.js/blob/61cbbb25a99e1ce995ddf0fac8571683c32e9b8b/src/access.js#L53
> 
> I wonder where is set coming from? Will this be assigned from the outside? Or is the implementation missing?

I found this commit that replaced `set` with `claim`: https://github.com/remotestorage/remotestorage.js/commit/b689d372b688caf8dab535df13bd4348cb10b20b

Looks like I overlooked that one to replace it as well.

johannesjo:
I am currently trying to fix the unit tests. Can you maybe help me understand this line:
https://github.com/remotestorage/remotestorage.js/blob/61cbbb25a99e1ce995ddf0fac8571683c32e9b8b/test/unit/localstorage-suite.js#L194
? 
Not really sure what that `fireInitial`does and where it comes from. Where are the different adapters mutated in such a fashion and why?

johannesjo:
Never mind. I am starting to understand how it works :)

galfert:
IMO it makes sense for `LocalStorage` to extend from `CachingLayer`. As for the event handling, a Mixin might be better suited. The decorator looks nice indeed.

skddc:
By the way, I just came across this blog post, containing some TS features I didn't know, but which might turn out to be helpful for refactoring our code (at whatever point in time):

https://thoughtbot.com/blog/going-through-changes-with-typescript

src/common-interfaces.ts
```diff
@@ -0,0 +1,24 @@
...
+  [key: string]: any;
+}
+
+export interface QueuedRequestResponse {
+  statusCode: number;
+  body?: object;
+  contentType?: ContentType
+}
+
+export type ContentType = any;
```
skddc, line 24: Shouldn't that always be a string?

src/localstorage.ts
```diff
@@ -1,7 +1,8 @@
-import * as cachingLayer from './cachinglayer';
 import * as log from './log';
 import * as eventHandling from './eventhandling';
 import {localStorageAvailable} from './util';
+import {RSNodes} from './common-interfaces';
+import {CachingLayerBase} from './cachinglayer-new';
```
skddc, line 5: I don't think we have to add the "Base" suffix there. Extending from something automatically makes the other thing a "base" thing.

skddc:
@johannesjo I'm actually seeing compilation errors like this for the latest version of this branch:

```
src/cachinglayer-new.ts:118:9 - error TS2349: This expression is not callable.
  Type 'typeof import("/home/basti/src/remotestorage/remotestorage.js/src/log")' has no call signatures.

118         log('[Cachinglayer] Error during PUT', nodes, e);
            ~~~
```

skddc:
Fixed the remaining SyncError test failure and also disabled linting in CI for now.

src/common-interfaces.ts
johannesjo: That's meant to be a placeholder until I figure out what ContentTypes we're expecting.

johannesjo:
> @johannesjo I'm actually seeing compilation errors like this for the latest version of this branch:
> 
> ```
> src/cachinglayer-new.ts:118:9 - error TS2349: This expression is not callable.
>   Type 'typeof import("/home/basti/src/remotestorage/remotestorage.js/src/log")' has no call signatures.
> 
> 118         log('[Cachinglayer] Error during PUT', nodes, e);
>             ~~~
> ```

I haven't found a good workaround to make it work with the current JavaScript imports while not throwing errors for  TypeScript. My suggestion would be to change up all imports and exports once everything is ported to typescript.

skddc:
We can still just use CommonJS modules like before. Works in TS as well. See `sync.ts` for example.

johannesjo:
> We can still just use CommonJS modules like before. Works in TS as well. See `sync.ts` for example.

Should I adjust all the existing occurences back to commonjs too?

skddc:
> Should I adjust all the existing occurences back to commonjs too?

Just anything that makes the compiler or tests fail, so we can work with a clean build.

johannesjo:
@skddc alright! Done!

skddc:
Yay, all green!

src/cachinglayer-new.ts
```diff
@@ -0,0 +1,433 @@
...
+        return Promise.all(calls).then(() => {
+          return allPaths;
+        });
+      });
+    } else {
+      return Promise.resolve([path]);
+    }
+  }
+
+  private _getInternals() {
```
iLiviu, line 344:
```suggestion
  _getInternals() {
```
_getInternals() is actually used in unit testing to give access to private methods even though it's notation indicates it's private. Wheter we should unit test private methods or not is another topic (most agree that only public methods should be tested, and test the private methods by calling the public methods with appropriate arguments).

src/cachinglayer-new.ts
```diff
@@ -0,0 +1,433 @@
...
+      }
+      return nodes;
+    });
+  }
+
+  flush(path: string): unknown {
+
+    return this._getAllDescendentPaths(path).then((paths) => {
+      return this.getNodes(paths);
+    }).then(function (nodes: RSNodes) {
```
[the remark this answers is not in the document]
skddc: > Also i think it would be nice to use `async/await` everywhere to make the code involving Promises cleanner and easier to understand

Totally agree. My fingers are yearning for the refactoring bonanza after the TS conversion. :)

skddc:
I just switched ESLint to work with TypeScript and check the TS files. Fixed a bunch of errors and some warnings in the process.

Warnings aren't breaking the build on CI, as they're ignored by `npm run lint`. There are now plenty of warnings left to improve the code bit by bit over time. But new code has to be error-free in this branch to keep it green now.

skddc:
I have moved the interfaces to global definition files and converted a few more modules.

The `Authorize` one was quite involving, and I also caught an existing, unreported bug on the way. See commit messages for details.

skddc:
Did some more things. See commit messages.

src/types.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  /**
+   * <uri> -> <alias>
+   */
+  aliases: { [key: string]: string } = {};
+
+  /**
+   * Called via public function BaseClient.declareType()
+   *
+   * @private (though not really)
```
skddc, line 41: The private in this JSDoc comment refers to the fact that it's not a public API for library consumers, not to being private within the library itself.

johannesjo: Yes. The comment was probably not ideal to make my intentions clear...

skddc:
Just added a to-do list to the PR description. Let me know if anyone sees something that's missing.

skddc:
Converted two more modules.

johannesjo:
Really not sure why the unit tests are failing for travis. It runs fine on my machine! ;)

I'll have a look a bit later. Sorry..

I just added the event handling as a mixin. There is a little bit of trickyness required to make it work with type completion, but I haven't found a more elegant approach to do it. 

@skddc I also recognized that we're using a little bit different approaches when it comes to migrating the old "classes". Keeping the prototype stuff has the advantage as we might be able to make the compiled code look more like before – though I suspect we can achieve similar results, by playing around with the "target" in the tsconfig.json – but to take full advantage of typescript, it might be advantageous migrating them to classes. At least for me it's quite helpful to get a grasp on which method is coming from where. 

Not sure if what's the best approach and when, if and how we should make some adjustments for consistency. What do you think?

johannesjo:
Just figured that the error is produced by the new unit tests introduced here:
https://github.com/remotestorage/remotestorage.js/commit/f1ab589c3f13565ac1e1c162b59faa01904864b7

@skddc Shall we merge/rebase master into our branch?

src/eventhandling-new.ts
```diff
@@ -0,0 +1,82 @@
+const log = require('./log');
+
+// wew use the mixin approach as described here:
+// https://mariusschulz.com/blog/mixin-classes-in-typescript
+type Constructor<T = {}> = new (...args: any[]) => T;
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type EventHandler = (event?: any) => void;
```
skddc, line 7: I think all `any` should at least throw warnings, because TypeScript is rather useless with `any` definitions in the chain.

src/eventhandling-new.ts
```diff
@@ -0,0 +1,82 @@
...
+
+export function eventHandlingMixedIn<TBase extends Constructor>(Base: TBase, additionalEvents: string[] = []) {
```
skddc, line 9: Is this a normal naming scheme for mixins in JS/TS? It sounds super weird to my ears.

src/inmemorystorage.ts
```diff
@@ -9,7 +8,7 @@ import {CachingLayer} from './cachinglayer-new';
  * @class
  **/
 
-export class InMemoryStorage extends CachingLayer {
+class InMemoryStorageBase extends CachingLayer {
```
skddc, line 11: What's the `Base` suffix for?

src/inmemorystorage.ts
```diff
@@ -39,17 +38,15 @@ export class InMemoryStorage extends CachingLayer {
...
+  protected static _rs_cleanup(): void {
     // empty
   }
 
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   private _storage: { [key: string]: any } = {};
 
   constructor() {
     super();
-    log('[InMemoryStorage] Registering events');
```
skddc: :+1:

src/inmemorystorage.ts
```diff
@@ -82,7 +79,9 @@ export class InMemoryStorage extends CachingLayer {
     return Promise.resolve();
   }
 
+  // NOTE: will be overwritten by eventHandlingMíxin
   _emit(...args): void {
```
skddc, line 83: A function that only throws should have a return type of `never`, which TypeScript actually knows and takes into account when the function is used then:

```suggestion
  _emit(...args): never {
```

However, I'm not sure if it's necessary to have this function included here in the first place.

src/localstorage.ts
```diff
@@ -1,7 +1,8 @@
+import {eventHandlingMixedIn} from './eventhandling-new';
+import {localStorageAvailable} from './util';
+import {CachingLayer} from './cachinglayer-new';
```
skddc, line 3: Just a formatting thing (which isn't linted at the moment): let's use spaces between curly braces and their content, same as before.

src/localstorage.ts
```diff
@@ -66,8 +65,9 @@ export default class LocalStorage extends CachingLayer {
 
 
   // TODO fix this
-  _emit (...args: any[]): void {
-    return;
+  // NOTE: will be overwritten by eventHandlingMíxin
+  _emit(...args: any[]): void {
```
skddc, line 69: Same as above: not sure if necessary. But if it is, then it should return as `never`.

skddc:
> Not sure if what's the best approach and when, if and how we should make some adjustments for consistency. What do you think?

As I said in my last comment, I think we should change as little as possible for now, because the changes won't be tracked in Git, and also we need to merge this branch asap, so we can actually go all-in on the refactoring asap.

skddc:
> I just added the event handling as a mixin. There is a little bit of trickyness required to make it work with type completion, but I haven't found a more elegant approach to do it.

I think we should refactor the event handling completely afterwards, and most likely put it in a basic feature class, which all other classes can inherit from. (That's just a wild guess for now; as soon as we have this branch merged, someone can come up with a proper design, that also works better with TypeScript.)

skddc:
> Just figured that the error is produced by the new unit tests introduced here:
> [f1ab589](https://github.com/remotestorage/remotestorage.js/commit/f1ab589c3f13565ac1e1c162b59faa01904864b7)
> 
> @skddc Shall we merge/rebase master into our branch?

GoogleDrive has already been ported to TS in this branch, so we'd need to apply the change in TS manually either way. But yeah, with a long-running branch like this, I'm afraid merging master from time to time is almost inevitable. :(

src/eventhandling-new.ts
johannesjo: It's unclear to me at this point how the events can look like, that's why I used any for now, but I agree that unknown is a better temporary solution.

I would suggest to spend some time when we converted all the files to typescript. There are a lot of declarations (e.g. the googledrive.ts) with no typescript type declared (even though they have jsdoc declarations). How would you feel about activating `noImplicitAny` flag in the tsconfig.json?

src/eventhandling-new.ts
johannesjo: Do you mean the camel case for `eventHandlingMixedIn`? It's not a real class but a more or less normal function, so I named it accordingly, but I am not personally attached to it :) What would you prefer?

src/inmemorystorage.ts
johannesjo: This is to avoid confusion on what is exported and what is to be used. Do you maybe have a better suggestion?

johannesjo:
> Converted two more modules. I started refraining from changing much code in the files, in favor of doing that with proper diffs later on. The sooner we get this merged, the sooner we can refactor in small PRs.

Sorry! I somehow missed that comment... But it makes a lot of sense to me!

> Edit: Forgot to mention that I think it's definitely useful to liberally leave TODO and FIXME comments all over the code, whenever you see something that should be refactored later. This way it doesn't get lost.

That's great. I like to use them even though not many of my former colleagues have been big fans of it :)

src/inmemorystorage.ts
skddc: I see. How about we just explain that requirement in the JSDoc comment then, and leave the name without the suffix?

src/eventhandling-new.ts
skddc: I mean the "MixedIn". I'd prefer to keep it to just to what it was, and document it in JSDoc if necessary. As you already gave thumbs-up on refactoring it completely, it won't last much longer anyway. :)

src/eventhandling-new.ts
skddc: > It's unclear to me at this point how the events can look like, that's why I used any for now, but I agree that unknown is a better temporary solution.

That's not what I meant. I meant it should throw a warning (instead of ignoring the error with this config), so we can later easily work down the linter warnings in the code. Maybe I misunderstood the code, and the warning is still intact with this comment?

skddc:
I fixed the build by simply adding the fix from master to the GDrive TS file, so that the test in master is green again.

src/wireclient.ts
```diff
@@ -71,7 +71,9 @@ if (typeof ((global || window as any).ArrayBufferView) === 'function') {
   };
 } else {
   const arrayBufferViews = [
+    // eslint-disable-next-line no-undef
```
skddc, line 74: I think these should be defined in the global eslint config here:

https://github.com/remotestorage/remotestorage.js/blob/feature/typescript_conversion/.eslintrc.js#L12-L19

johannesjo:
@skddc just came across this:
https://www.sitepoint.com/getting-started-with-codemods/

Probably all not too useful in the course of this refactoring, but pretty interesting to me nonetheless. Haven't tested anything so far, but intend to give it a spin. There is also a (remote) [meetup](https://www.meetup.com/de-DE/gdg-berlin/events/270067784/) next week about this subject hosted by our corporate overlords in case youre interested.

skddc:
The Ember.js community has been using codemods for framework upgrades for a long time now. Most of my apps are based on Ember, so I use them regularly.

As you say yourself, I don't think they'll be overly useful for this codebase, because refactoring it needs too much human oversight in my opinion. But hey, if it's fun and interesting to you, try it out here as well! :)

You can also use the forums for any non-issue/PR-specific conversations: https://community.remotestorage.io/c/remotestorage-js-core/8

src/eventhandling-new.ts
johannesjo: Ahhh. Sorry, I misunderstood you. I added a todo on top. I leave it at that for now.

johannesjo:
Hey! Just wanted to let you know that I decided that I am likely to need something else for my needs (probably a combination of automerge and blockstack or maybe maybe solid). Because of this it doesn't make much sense for me to continue working on this Pull Request or the specification for the to do model. I  hope you understand.

skddc:
That's odd, but no problem. Good luck with whatever you end up choosing!

(Would be nice if you could let us know what RS is missing in terms of functionality somewhere outside of this PR.)

johannesjo:
It's a combination of many reasons. But first of all what I want to achieve is much easier with Blockstack, as I need to do a lot less adjustments to my code and thus to commit much less to the platform. The same is only partly true for automerge but in turn it gives me a much stronger conflict resolution.
 
Another reason is that RemoteStorage is not just a library but a platform. If I were to commit to it I would be bound to it long term. Blockstack, Automerge & Solid all seem to have an edge in this area, more people seem to be using it and the development is much more active in the main libraries and equally important in the stuff around them.

skddc:
I really don't want to add more unrelated comments to this PR, so it's a bit sad that we can't talk about it elsewhere (especially because I don't understand how half of it even relates). Thanks anyway for adding more info about your perspective! Much appreciated.

skddc:
I have just finished porting the rest of the source files, and I will continue with the remaining tasks next week.

skddc:
Status update: I have pushed a version that is all TypeScript now, and which also uses ES module exports and imports everywhere in the source files. It's probably easier to examine the changes by looking at the list of commits, because I had to change and refactor quite a few things.

One of the commits removes the JavaScript files completely from source. They are compiled to a (git-ignored) `/build` folder now, from where the tests will import them, and from which Webpack can build releases. (I'm going to document all of the new setup properly in the "Contributing" section of the docs soon.)

I'm not sure why the Travis build is failing for this last commit. Would be nice if someone could run `npm test` on their machine and let me know if it's a general problem or not. The general code changes are also ready for review now. I would only adjust the build/release scripts a bit more, if necessary. More refactorings and improvements can then be done in small PRs, so it's much easier to review and also work on it parallel.

iLiviu:
Travis build is failing because the `test/unit/build ` symlink that i think you created is not tracked by git due to the `.gitignore` rules (should change it to only ignore parent `/build` directory?).

skddc:
> Travis build is failing because the `test/unit/build ` symlink that i think you created is not tracked by git due to the `.gitignore` rules (should change it to only ignore parent `/build` directory?).

Ah yes, stupid me. Thanks!

skddc:
I forgot to mention: there's also a problem with unhandled rejected promises in a few tests of WireClient, but due to how the tests are written it's basically impossible to know which ones are failing. I tried to find the actual bug for almost 2 hours, but failed. I then had a call with @silverbucket who said we may have to change the tests to return the promises so they can be detected as async tests and fail properly.

skddc:
All green now! Looking forward to your reviews.

silverbucket:
Updated the tests to catch those pesky async exceptions in WireClient! All tests pass without any exceptions thrown.

skddc:
Fantastic, thanks!

skddc:
So, for the last couple of days I wrestled with sphinx-js, because for some reason it can't find functions via `autofunction` anymore. I tried it with @silverbucket, too, and it was the same on his machine (with different OS and different Sphinx 2 vs. 3).

Until today, I thought this only happened since I switched from JSDoc to TypeDoc, but I just double-checked the current master, and the exact same thing actually happens for our current JSDoc setup as well. :/

The errors look like this:

    No JSDoc documentation was found for object "Access#claim" or any path ending with that

Unfortunately, there's no backtrace, or anything else that would help with debugging. I also tried using path syntax, instead of the capitalized class name syntax for finding the function, but no dice. And `autoclass` also doesn't find anything. If anyone has experience with Sphinx or Python, any help would be appreciated. (From current sphinx-js issues, it doesn't look like it's a common problem for other users.)

skddc:
Would be cool if some people could review/test the current state of this branch (minus the docs). Thanks!

galfert:
I can confirm that the test suite is green for me locally. So far I didn't have much time to have a deeper look at this PR, but I hope I will get to it in the next few days.

src/features.ts
```diff
@@ -131,11 +144,9 @@ const Features = {
   },
 
   featureFailed (featureName, err) {
-    log(`[RemoteStorage] [FEATURE ${featureName}] initialization failed (${err})`);
-    this.featureDone();
+    throw new Error(`Feature initialization failed for "${featureName}": ${err}`);
```
galfert, line 147: I'm not sure about this change. Wouldn't the whole library initialization fail this way, instead of one particular feature that would be missing? E.g. when IndexedDB initialization failed, the whole library wouldn't work instead of just falling back to localStorage.

skddc: Good catch! I think you're right.

skddc:
Status update:

Notwithstanding the fact that we saw the same bug in Sphinx 2.8, the developers have released a version of sphinx-js that is officially compatible with Sphinx 3 in the meantime. There are also a bunch of open PRs on the sphinx-js repo, some of which sound relevant to the bug that we're seeing.

I pushed the updated version, as well as the TypeDoc config and contributing doc to this branch now. Unfortunately, I'm now seeing a different (similarly cryptic) bug that prevents building the docs:

```
Extension error:
Handler <function gather_doclets at 0x7f0d9d2519d0> for event 'builder-inited' threw an exception
```

---

On a side note, they're also working on decoupling sphinx-js from JSDoc more, so that it's possible to add more TypeScript (and other) features soon. Sounds good to me.

skddc:
Thinking about this some more, I changed my opinion about the status of this PR.

How about we merge it as is, and document the remaining tasks in separate issues? This way we can start refactoring the TS code at the same time as resolving the sphinx-js bugs and preparing the final release process. And we can better split those tasks as well.

We would just not be able to tag a release from master until the blockers have been resolved. But we already have a blocker label for this purpose.

What do you think?

galfert:
I don't have any objections to this.

skddc:
Thanks.

Just a quick status update: After fixing some last bugs I found with the actual release build, I'm currently in JavaScript module hell, trying to get the release and test suite to work at the same time. I have switched building to be all done by Webpack directly, so we also get TS stacktraces in browsers even. But I have to figure out some last things before I can push my commits for this.

skddc:
Alright, I have added new issues for the task that remain before we can tag a release after merging this. I have also added a milestone for it:

https://github.com/remotestorage/remotestorage.js/milestone/16

Any and all help with testing, as well as adding, commenting on, or solving issues in that milestone, would be most appreciated, of course.

Issue #1178, opened 2020-04-11
Dropbox Provider does not Preserve Case


kevincox:
It appears that the `list_folder` HTTP API is used and the code uses the `path_lower` property instead of the `path_display` property.

DougReeder:
This causes files with capital letters in their path to appear twice in listings, which is a severe problem.

PR #1179, opened 2020-04-11
Fix case preserving for Dropbox

kevincox:
Untested fix for https://github.com/remotestorage/remotestorage.js/issues/1178

### DIFF
--- a/src/dropbox.js
+++ b/src/dropbox.js
@@ -221,7 +221,7 @@ Dropbox.prototype = {
 
       listing = body.entries.reduce(function (map, item) {
         var isDir = item['.tag'] === 'folder';
-        var itemName = item.path_lower.split('/').slice(-1)[0] + (isDir ? '/' : '');
+        var itemName = item.path_display.split('/').slice(-1)[0] + (isDir ? '/' : '');
         if (isDir){
           map[itemName] = { ETag: revCache.get(path+itemName) };
         } else {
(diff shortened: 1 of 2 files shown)


skddc:
The tasks left that I see:

* Check if all supported APIs/protocols preserve case (especially RS). If not, then preserving case would be incompatible between them, thus impacting data portability.
* Add a test for preserving case specifically

I didn't find anything about case sensitivity in the RS spec at first glance. Perhaps @michielbdejong or @fkooman have insight on this?

I also didn't find a test for it in the [RS API test suite](https://github.com/remotestorage/api-test-suite/), which means it's probably unspecified and different servers may treat it differently at the moment.

skddc:
@michielbdejong Have you seen the question here by any chance?

DougReeder (changes requested):
This approach will fix #1178, which is severe. There will remain some edge cases which are not handled gracefully. Fully handling case-insensitivity would appear to require changes to the sync engine, which deserve a separate commit.

(These commits should probably be cherry-picked into a new PR.)

raucao:
@DougReeder Is the "changes requested" review status about cherry-picking the commits into a new branch?

DougReeder:
Yes, this just needs to be rebased or cherry-picked into a new branch, and to check that all tests still pass.

raucao:
New PR: #1275

PR #1181, opened 2020-04-14
Handle Google Drive paths with special characters

### LINKED ISSUE #1176: Google Drive Folders with "special characters" get duplicated.
Possibly related to: https://github.com/remotestorage/remotestorage.js/issues/682

Example code:
```html
<body>
<script src="https://unpkg.com/remotestoragejs@1.2.3/release/remotestorage.js"></script>
<script src="https://unpkg.com/remotestorage-widget@1.5.1/build/widget.js"></script>
<script>
const remoteStorage = new RemoteStorage();
remoteStorage.setApiKeys({
	googledrive: "$apikey",
});

setTimeout(() => new Widget(remoteStorage).attach(), 1000);

remoteStorage.access.claim("ca.kevincox.example", "rw");

remoteStorage.on("connected", () => {
	let scope = remoteStorage.scope("/ca.kevincox.example/");
	// Works
	// scope.storeFile("text/plain", "file", "foobar");
	// Loops forever
	scope.storeFile("text/plain", "some folder/file.json", "");
});
</script>
```

The file `file.json` never gets wrtten and there are constant network requests creating new folders called "some folder".

iLiviu:
Paths were not encoded when used as keys for Id cache to insert data,
but when data was retrieved from cache, the paths were used encoded,
so folders with special characters were never identified as already
existing.

Fixes #1176

### DIFF
--- a/src/googledrive.js
+++ b/src/googledrive.js
@@ -524,12 +524,12 @@ GoogleDrive.prototype = {
         for (const item of data.items) {
           etagWithoutQuotes = removeQuotes(item.etag);
           if (item.mimeType === GD_DIR_MIME_TYPE) {
-            this._fileIdCache.set(path + item.title + '/', item.id);
+            this._fileIdCache.set(path + cleanPath(item.title) + '/', item.id);
             itemsMap[item.title + '/'] = {
               ETag: etagWithoutQuotes
             };
           } else {
-            this._fileIdCache.set(path + item.title, item.id);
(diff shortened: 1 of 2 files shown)


skddc:
LGTM :+1: 

Should we maybe add a test with special characters?

iLiviu:
I added a test which fails on master, but passes after applying current fix.

skddc:
Perfect!

PR #1193, opened 2020-08-04
WIP: Port tests to Mocha and Chai

raucao:
Jaribu is a bit dated now, and it's still using AMD modules. I talked to @silverbucket (the author of Jaribu, and a rs.js contributor) today, and he agrees that we should eventually switch to a more modern, widely used framework.

So, this sets up Mocha as a test runner and Chai for assertions, as well as ES modules/imports. A first working test file is included.

The new tests end in `.test.js` instead of `-suite.js`, and they can be run via `npm run test:mocha` for now.

Standard output looks like this (but can be changed to other formats via CLI arguments):

Screenshot from 2020-08-04 22-14-35


skddc:
... I also talked to @galfert about this earlier. As it's a *ton* of code to port, we agreed that it makes sense to share this work.

I also threw in the idea of merging chunks of ported files before everything is finished, to make reviews easier and work in parallel even more efficient. (We could just add `test:mocha` to CI, in addition to Jaribu, until there are no Jaribu files anymore). Not sure about this yet. Opinions and suggestions welcome!

skddc:
I should have looked into using TypeScript directly in the tests, of course. It is possible with `ts-node`...

skddc:
Got it working! See latest commit. (`npm run test:watch` FTW!)

Already found a hidden bug (just unnecessary test code) this way, because TypeScript knew it wasn't a valid property on that class.

silverbucket:
> ... I also talked to @galfert about this earlier. As it's a _ton_ of code to port, we agreed that it makes sense to share this work. Any and all help is much appreciated! Just drop a quick comment here for which files you're going to port, before you start.
> 
> I also threw in the idea of merging chunks of ported files before everything is finished, to make reviews easier and work in parallel even more efficient. (We could just add `test:mocha` to CI, in addition to Jaribu, until there are no Jaribu files anymore). Not sure about this yet. Opinions and suggestions welcome!

Yeah this is a sound approach, and what I'm doing in Sockethub as well. There shouldn't be any issues having two test runners running back to back until all the existing tests are ported from Jaribu to Mocha.

I'll try to jump in and help as well when I have the time. Perhaps it would be best to merge the PR which contains all of the config and setup (with both test runners executed as part of unit tests) then we can submit PRs as we tackle specific files?

raucao:
> Perhaps it would be best to merge the PR which contains all of the config and setup (with both test runners executed as part of unit tests) then we can submit PRs as we tackle specific files?

I think so, yes. Will take a look at the merge conflict within the next few days.

DougReeder:
Excellent! We should consider using https://www.npmjs.com/package/eslint-plugin-mocha

I support merging this as soon as it's ready for others to use it to write tests.

DougReeder:
An independent but related question - do we want to continue to support testing with XHR?  XHR really complicates mocking the network for tests.

DougReeder:
I've merged master in master and updated the dependencies to the latest.  All the tests pass under Node v14 and v16, but the linter is complaining about Typescript issues that don't mean much to me.  Someone with a better knowledge of Typescript should sort them out.

As the jaribu tests fail under Node v18, and Node v16 will go EOL in one year (2023-09-11), we need to move this forward.

DougReeder:
I've fixed the eslint configuration, and this looks ready to merge, to me.  I'd very much like another dev to try it out, first.

DougReeder:
I'm adding chai-as-promised to handle async and sinon to handle mocking/spies in the Dropbox tests (in another branch).  If you'd rather we standardized on something else, please let me know right away.

raucao:
Since this has not been cleanly rebased on top of master, but master instead merged into the branch, it's not really feasible for me to see why there are issues with both running `npm run dev` as well as transpiling the `@types` modules using `tsc` now.

The branch did not have any such issues before the merge. I will rebase it on top of master to see what's going on.

There is also #1266 with a whole bunch of TypeScript-related refactorings, just FYI. Not sure if it's better to merge that first.

raucao:
Continued in #1271.

PR #1196, opened 2020-09-02
Type definition publishing

### LINKED ISSUE #1190: Implement process for publishing type definitions
The `build:js` npm script currently puts TS definition files in `release/types`.

We need to decide if that's where we want the definitions to live, and if we need/should to publish them to a `@types` npm repo, in addition to whatever we can do in `package.json` for auto discovery.

skddc:
* Add the "types" property to `package.json`, so consumers can find the definition files
* Add source maps for the type definitions, so consumers can find the original TypeScript source code
* More (necessary) TypeScript porting of source files

closes #1190 

---

For future reference: https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html


skddc:
I had to refactor the `RemoteStorage` and `Dropbox` legacy constructor functions into classes, and then fix a bunch of errors as well. Also, the `types` property in `package.json` needs a file instead of a directory.

I have pushed a very bare-bones Ember/TypeScript test app where everything works now:

https://gitea.kosmos.org/raucao/rs-ember-typescript-test

You can just run `npm i && npm start`, and then open your browser console while using it. You can `remoteStorage.connect('user@host')` from the console, and it should sync your favorite drinks. ;)

galfert:
When running the test app, I see two errors:

```
node_modules/remotestoragejs/release/types/caching.d.ts:1:33 - error TS2307: Cannot find module './interfaces/caching'.

1 import { CachingStrategy } from './interfaces/caching';
                                  ~~~~~~~~~~~~~~~~~~~~~~

app/services/storage.ts:8:3 - error TS2564: Property 'remoteStorage' has no initializer and is not definitely assigned in the constructor.

8   remoteStorage: RemoteStorage;
    ~~~~~~~~~~~~~
```

The build doesn't stop though and the app seems to work.

skddc:
> EDIT: I think the second error can be fixed by renaming the method `initialize()` to `constructor()` in `app/services/storage.ts` (and adding a `super()` call in the first line of the constructor). And removing the complete `beforeModel` block from `app/routes/application.ts`.

That's what I had, but it didn't execute the constructor code at all. I think the error is coming from the Ember TS add-on being a bit of a hack.

> `node_modules/remotestoragejs/release/types/caching.d.ts:1:33 - error TS2307: Cannot find module './interfaces/caching'.`

This one I haven't seen before. Does caching work when you connect an account and sync some drinks?

skddc:
@galfert I have fixed the type declaration import issue.

galfert:
I can confirm that the error is gone.

skddc:
> I can confirm that the error is gone.

Does that mean you approve the PR or is something else missing?

galfert:
I don't know enough about TS type definitions to approve this. But I don't have any objections.

skddc:
OK, thanks.

PR #1199, opened 2020-09-22
Fix SphinxJS/TypeDoc integration

### LINKED ISSUE #1192: Fix sphinx-js integration
We recently ran into sphinx-js bugs. See comments in #1175 for details, and mozilla/sphinx-js#147 for the open issue on the source repo.

skddc:
The new version of SphinxJS has been released, and it has official TypeDoc support now. I successfully got it running on my machine. But I had to refactor the last remaining non-native classes/functions for the autofunction lookups to work.

closes #1192

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,4 +9,5 @@ server/server-state
 server-state
 doc/code
 doc/_build
+doc/__pycache__/
 /remotestorage*.js
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,6 @@
+version: 2
+
+python:
+  version: 3.8
+  install:
+    - requirements: docs/requirements.txt
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -3,4 +3,4 @@ version: 2
 python:
   version: 3.8
   install:
-    - requirements: docs/requirements.txt
+    - requirements: doc/requirements.txt
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -184,3 +184,10 @@ texinfo_documents = [
      author, 'remoteStoragejs', 'One line description of project.',
      'Miscellaneous'),
 ]
+
+#
+# HACKFIX WARNING
+# TODO Remove this when there is official support for pre-build steps on RTD
+# https://github.com/readthedocs/readthedocs.org/issues/6662
+#
+os.system('npm install')
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -43,7 +43,7 @@ templates_path = ['_templates']
 
 # Used to extract JSDoc function/class docs from source
 js_language = 'typescript'
-js_source_path = '../src'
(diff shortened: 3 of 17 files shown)


skddc:
Something is still wrong with ReadTheDocs' TypeScript/TypeDoc: https://readthedocs.org/projects/remotestoragejs/builds/11943970/

It's missing our npm type dependencies apparently, and it looks like it's also using outdated TS. I'm having a look.

skddc:
I couldn't make it work so far, but spent quite some time already. Looking at [a raw build log](https://readthedocs.org/api/v2/build/11945380.txt), it seems to me like RTD is not actually doing `npm i` during the build process, so we'd have to check in and push the `node_modules` folder.

Improved some other things on the way, like e.g. having more than just master/latest being published. `v1.2.3` is now published explicitly in addition, and used as default redirect. Visitors can then switch to "latest" if they choose so.

skddc:
Eric from RTD got back to me and suggested to just hackfix the pre-build step by shelling out from `conf.py`, which is a Python file that is being executed before the build. So I tried that, and it worked immediately:

https://remotestoragejs--1199.org.readthedocs.build/en/1199/contributing/docs.html

Everything ready to merge now, and it will publish the new docs as "latest", as soon as this PR lands in master.

galfert:
Documentation builds are working for me.

For the autobuilding I first tried running `npm run doc`, because that's the first task I found in the `package.json`.

But that task faild with an error:

```
sphinx-autobuild: error: unrecognized arguments: -i -i -i #*# . _build/html
```

Looks like the shorthand version for the `--ignore` argument isn't supported anymore by `sphinx-autobuild`.

But then I noticed that the documentation documentation (doc/contributing/docs.rst) mentions the task `autobuild-docs` instead. It's the same as the `doc` task, just without the ignore arguments. So I removed the `doc` task to prevent any further confusion.

In the doc builds themselves I noticed that all number arguments have the type `unknown` now:

Screen Shot 2020-09-24 at 14 04 43

But I guess that can be fixed as part of #1201.

skddc:
> In the doc builds themselves I noticed that all number arguments have the type `unknown` now:

We'll have to find a solution for that, because the type is actually supposed to be "unknown" in the code, so that TypeScript forces us to validate the input from the app author. But yeah, I agree that we can handle that later.

skddc:
I couldn't find a solution for the `unknown` type problem, so I created an issue about it here:

https://github.com/TypeStrong/typedoc/issues/1370

skddc:
Talked about it with @galfert and we came to the conclusion that using normal types instead of `unknown` is the best solution here. Losing the type safety isn't worth the effort here.

skddc:
I just pushed some commits to address the last documentation issue (`unknown` argument types). Ready to merge AFAICS.

galfert:
Looks good to me 👍

PR #1207, opened 2020-10-22
WIP: Fix ReadTheDocs npm-install issue

raucao:

PR #1210, opened 2020-12-09
Fix WireClient initialization

galfert:
This fixes an error thrown during library initialization:

> TypeError: this.remote is undefined

It happens when there are no settings for the WireClient in localStorage (i.e. userAddress, token, ...), preventing the WireClient instance to be created. The error that is (silently) thrown is

>  TypeError: Cannot destructure property 'userAddress' of 'util.getJSONFromLocalStorage(...)' as it is null.

### DIFF
--- a/src/wireclient.ts
+++ b/src/wireclient.ts
@@ -63,7 +63,9 @@ const STORAGE_APIS = {
 };
 
 function readSettings () {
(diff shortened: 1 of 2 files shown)


silverbucket (approved):

silverbucket:
This LGTM but it looks like we have some doc errors

galfert:
> looks like we have some doc errors

I think that's a known issue that is being worked on in #1207.

raucao (approved):
:+1:

Issue #1191, opened 2020-08-04
Fix/update node.js usage/integration


raucao:
The TS port (#1175) was merged to master without caring for node.js usage yet. Should be easy to make it work, as we already have a separate entry point for node.js usage (`index.js`). We can probably use https://www.npmjs.com/package/esm for that.

skddc:
I just tagged `2.0.0-alpha.2`, which I successfully used in a node.js program. However, it is currently necessary to access the constructor via `RemoteStorage.default` when requiring it as a CommonJS module. And I'm wondering if the package mentioned above could help with making that nicer for node users.

Then again, RS modules also have to be loaded with `.default` already, so maybe having it be more consistent is better, and maybe nudging node users towards ES modules is also not a bad idea.

Anyone has opinions on this? Theoretically, this could be closed with just a minor documentation change now (about requiring an AJAX or Fetch lib).

raucao:
Just tested this with a node program that was using the current stable release. No changes necessary anymore, and you get code completion with integrated docs, if you have a TS language server working in your text editor. \o/

PR #1211, opened 2020-12-09
Remove usage of require().default when using RemoteStorage in node

galfert:
Refs #1191

I found a feature of TypeScript that prevents needing to use `require().default`
in node, when a module only has a single export.

Instead of `export default Foo` you can just use `export = Foo`.


raucao (approved):
Fantastic! :clap:

PR #1213, opened 2020-12-22
Don't add trailing slash as only pathname to redirect_uri param

### LINKED ISSUE #1212: GoogleDrive `redirect_uri=http://localhost:port/` should be `redirect_uri=http://localhost:port`
0. Open console `https://console.developers.google.com/apis/credentials/oauthclient/${your_client_id}?project=${your_project_number}`
1. Set `http://localhost:port` as allowed to redirect and scripts. (URL ends with "/" - not valid)
2. Go to `http://localhost:port`.
3. Select GoogleDrive in Widget.
4. App sends request like: `https://accounts.google.com/o/oauth2/auth?redirect_uri=http://localhost:port/&scope=https://www.googleapis.com/auth/drive&client_id=...&response_type=token`
5. Oops.
```
Error:
The redirect URI in the request, http://localhost:port/, does not match the ones authorized for the OAuth client.
```

P.s. If i send request manualy without "/" it works fine.

env:
    "remotestorage-widget": "^1.5.2",
    "remotestoragejs": "^2.0.0-alpha.5",

galfert:
Fixes #1212

When the current URL is the root of the host (e.g. https://example.com), don't add a trailing slash (i.e. https://example.com/) to the `redirect_uri` param when sending the user to the OAuth provider.

### DIFF
--- a/src/remotestorage.ts
+++ b/src/remotestorage.ts
@@ -257,7 +257,17 @@ class RemoteStorage {
       options.scope = this.access.scopeParameter;
     }
 
-    options.redirectUri = globalContext.cordova ? config.cordovaRedirectUri : String(Authorize.getLocation());
+    if (globalContext.cordova) {
+      options.redirectUri = config.cordovaRedirectUri;
+    } else {
+      let location = Authorize.getLocation();
+      let redirectUri = location.origin;
+      if (location.pathname !== '/') {
+        redirectUri += location.pathname;
+      }
+
+      options.redirectUri = redirectUri;
+    }
 
     if (typeof options.clientId === 'undefined') {
       options.clientId = options.redirectUri.match(/^(https?:\/\/[^/]+)/)[0];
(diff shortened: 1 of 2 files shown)


src/remotestorage.ts
```diff
@@ -257,7 +257,17 @@ class RemoteStorage {
       options.scope = this.access.scopeParameter;
     }
 
-    options.redirectUri = globalContext.cordova ? config.cordovaRedirectUri : String(Authorize.getLocation());
+    if (globalContext.cordova) {
+      options.redirectUri = config.cordovaRedirectUri;
+    } else {
+      let location = Authorize.getLocation();
```
raucao, line 263: I think this should be `const`, as it's not redefined, no?

raucao (approved):
Nice! :+1: 

Just one nitpick.

src/remotestorage.ts
galfert: Right.

PR #1214, opened 2020-12-22
Generate separate sourcemap instead of inlining it

galfert:
Inlining the sourcemap in the build file increases the filesize quite a lot.

This changes it back to a separate file as it was in 1.x.


raucao:
So sorry. :)

raucao (approved):
Yes! :clown_face:

PR #1215, opened 2021-01-12
Throw error instead of string

rosano:
I am not able to `catch` this error in mocha because it's a string. Might be useful to do this throughout the code but I don't know if there will be side effects. Tests seem to pass with this change.


raucao:
Makes sense. Thanks!

PR #1217, opened 2021-02-13
Fix typo

rosano:
The slashes that are sent with `scope` are invalid for `claim`.

### DIFF
--- a/src/remotestorage.ts
+++ b/src/remotestorage.ts
@@ -694,7 +694,8 @@ class RemoteStorage {
     }
 
     if (!this.access.checkPathPermission(path, 'r')) {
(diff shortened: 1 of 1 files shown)


raucao:
> The slashes that are sent with `scope`

I'm a bit unclear as to where they are "sent". From the code change it looks like printed output is using extra characters around a path string?

rosano:
Following this example from the docs `const client = remoteStorage.scope('/foo/');` the slashes are part of the argument?

raucao:
Had a look playing with the actual code just now.

Stripping the first and last character would also create an invalid command when someone uses a wrong path argument. So I fixed it in a different way in #1218 and added you as a co-author. Could you review that one, @rosano?

PR #1218, opened 2021-02-15
Fix potentially wrong command in warning message

raucao:
Improves the warning message by removing the exact command entirely and referring to the proper documentation instead.

Replaces #1217.

Screenshot from 2021-02-15 11-44-37


rosano:
Much better. I guess that `escapedPath` variable could also be eliminated then.

raucao:
>  I guess that `escapedPath` variable could also be eliminated then.

Of course! Thank you.

Issue #1222, opened 2021-03-07
Discussion for switiching from travis-ci.org to ? for continuous integration


pjbollinger:
While working on this project, I noticed there is the use of travis-ci.org for performing the continuous integration. It has been announced that travis-ci.org will be shutting down in favor for travis-ci.com. The shutdown was supposed to be at the end of 2020, but they seem to be holding off for now. https://docs.travis-ci.com/user/migrate/open-source-repository-migration/#frequently-asked-questions

My concern is the project is on living on borrowed time when it comes to the continuous integration tool. I feel there are a few options available to address this and ensure no blocks in the future:
* Switch to travis-ci.com, which will still be free per https://docs.travis-ci.com/user/migrate/open-source-repository-migration/#q-will-travis-ci-be-getting-rid-of-free-users
* Switch to GitHub Actions, now that GitHub has free built-in CI for open source
* Switch to another system like CircleCI

Personally, I'd be inclined to use GitHub Actions since it is gaining a lot of traction so the feature should be supported for a long time. If you'd like, I can create a new PR that implements GitHub Actions as a proof-of-concept.

Thoughts?

raucao:
Personally, I'm fine with switching to anything that works, and that contributors like. Locking more things in to a Microsoft service feels a bit icky to me, but self-hosting a CI server is probably not worth it for us either. And in the end, CI builds based on Docker images aren't really a pain to potentially migrate to other platforms in the future anyway.

@remotestorage/core Anyone else has opinions on this? If nobody objects before tomorrow, I'd suggest to go ahead with it as proposed by @pjbollinger.

pjbollinger:
I can understand the sentiment about Microsoft 😆 

I feel the easiest option would be to switch to travis-ci.com but it would require intervention from the repository owner(s)/admin(s): https://docs.travis-ci.com/user/migrate/open-source-repository-migration

galfert:
Hi. First of all, thank you @pjbollinger for helping with this. Personally I don't have any objections to any of the mentioned options.

michielbdejong:
@pjbollinger thanks for bringing this up! Most other projects I'm involved in switched to GitHub Actions, too.

> Personally, I'd be inclined to use GitHub Actions since it is gaining a lot of traction so the feature should be supported for a long time. If you'd like, I can create a new PR that implements GitHub Actions as a proof-of-concept.

Great!

[...]

PR #1223, opened 2021-03-08
Add a GitHub Actions Workflow

pjbollinger:
This GitHub Action Workflow should replicate the continuous integration that .travis.yml/travis-ci.org performed. This is in regards to #1222.

If the project is happy with how this runs and after a little time, a new PR can be created to remove Travis CI to close the referenced issue. We could remove Travis CI now but there's no harm in running both for a bit to verify functionality.

### DIFF
--- a/.github/workflows/test-and-lint.yml
+++ b/.github/workflows/test-and-lint.yml
@@ -1,0 +1,27 @@
+name: test-and-lint
+on:
+  push:
+    branches: [ master, stable ]
+  pull_request:
+    branches: [ master, stable ]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # Support LTS versions based on https://nodejs.org/en/about/releases/
+        node-version: ['10', '12', '14']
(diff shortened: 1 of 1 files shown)


raucao:
Great! I just enabled "local actions" for this repo, but no idea how to activate or test this workflow now.

pjbollinger:
If you're referring to this doc, https://docs.github.com/en/github/administering-a-repository/disabling-or-limiting-github-actions-for-a-repository#allowing-specific-actions-to-run, I don't think `Allow only local actions` is the right setting as it won't let the repository run stuff from GitHub's `actions` repository.

I can understand not wanting to allow all Actions, so perhaps a better option would be to select `Allow select actions` and then ensure `Allow actions created by GitHub` is checked.

I'm not sure but the GitHub Actions might not run until the PR is merged since it's not part of the repository yet.

raucao:
OK, cool. I enabled the select actions, both from GitHub and verified creators on Marketplace. Let's see what happens after merging...

PR #1224, opened 2021-03-08
Fix GitHub Actions Workflow

pjbollinger:
Testing this locally with https://github.com/nektos/act, I was able to reproduce the issue and fix it with this change.
Now that the workflow has been committed to the repository, hopefully it will be easier to test and check.

Apologies for the additional PR, I should have looked to see how to test it locally before committing. 😅

### DIFF
--- a/.github/workflows/test-and-lint.yml
+++ b/.github/workflows/test-and-lint.yml
@@ -19,8 +19,6 @@ jobs:
           node-version: ${{ matrix.node-version }}
       - name: Install dependencies
         run: npm install
-      - name: Compile TypeScript
-        run: tsc
       - name: Run tests
         run: npm test
       - name: Run linter


.github/workflows/test-and-lint.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       - name: Install dependencies
         run: npm install
       - name: Compile TypeScript
-        run: tsc
+        run: npx tsc
```
raucao, line 23: The npm test script actually includes compiling TypeScript already, so I think this step is not necessary.

pjbollinger: I agree 👍  I wasn't sure exactly why this separate `tsc` was called out in the first place, other than a sanity check, but I'll go ahead and remove this step.

raucao: Yeah, just saw that it was already unnecessary on Travis. Just an oversight when we were working on the whole TS port I guess.

raucao:
:tada:

PR #1227, opened 2021-03-09
Fix/improve bash snippet formatting

raucao:
The command snippets in the docs are currently marked as either `bash` or nothing (i.e. automatic formatting). However, there's a better way, which highlights the `$` prefix, as well as prevents it from being copied when selecting the entire line to copy and paste it.

Edit: selecting only the command itself, not including the '$` prefix, only works with the build on my local machine, but not in the RTD preview build so far. So maybe we'll have to wait for them to update their theme in production for this to work.

### DIFF
--- a/doc/contributing/building.rst
+++ b/doc/contributing/building.rst
@@ -27,7 +27,7 @@ Production
 
 .. CODE:: console
 
-   $ npm run build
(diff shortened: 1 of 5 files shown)


galfert (approved):
Looks all good to me 👍

PR #1230, opened 2021-03-10
Only install Node.js dependencies if missing for docs build

### LINKED ISSUE #1229: Don't run `npm install` on every local doc build
It's even run when rebuilding the docs automatically with `autobuild`. It's part of the Sphinx/Python config file, and is needed to install our deps on RTD:

https://github.com/remotestorage/remotestorage.js/blob/ae30441bf1cd2c348592cd1cb326f5847bcbdbfb/doc/conf.py#L208

pjbollinger:
Currently, every time the docs are built we run `npm install`.
This can be frustrating for developers because using `autobuild` for the docs leads to `npm install` constantly being run.
By testing if we have a local copy of `typedoc`, we can avoid the need to always re-run `npm install`.
This fixes #1229.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -18,6 +18,7 @@
 import datetime
 import os
 import sys
+from pathlib import Path
 sys.path.insert(0, os.path.abspath('.'))
 
 from version import __version__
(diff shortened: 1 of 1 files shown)


raucao:
Great! :+1:

PR #1232, opened 2021-03-12
Don't include unnecessary files in npm package

raucao:
... plus some general housekeeping. See commit messages.


galfert (approved):
LGTM

Issue #1170, opened 2019-11-04
connect() overwrites local data and disconnect() clears local data, why?


dienluong:
Hello,

First, thank you for the project!

rs.js is advertised as "offline-first by design".  But how can an app be offline-first when local data is not kept?

To give a simple example: let's say we have a notes taking app.  When not connected to a remote storage, the user can still create, delete and update notes.  Once connected to online storage, the new or modified notes get sync'd with the data in online storage.  That's not possible with rs.js because, based on my testing, when you connect, the online version of your data overwrites the local copy (i.e. your local modifications are lost).  Furthermore, when the user disconnects from the online storage, the local copy is destroyed.  So the user can no longer work on the existing notes, when disconnected.

So it seems to me that rs.js, despite the claimed "offline-first design", does not enable an app to be offline-frst.

Update: I think I realize now what "offline-first design" means in the context of rs.js.  It means no Internet/remote connection but still "**logged in**" to a remote storage account (as opposed being in "anonymous mode").

skddc:
Hi Dien,

Thanks for your question. The answer to this is two-fold, and after testing this for a bit in one of my apps, it seems you may have found a bug in the library. Let me explain:

### Connecting a remote storage with existing local data

Data created or edited while being offline/disconnected should *not* overwrite remote data by default. When connecting a remote storage, the data should be synced exactly the same way as if your app had been connected, but temporarily lost Internet connectivity.

The way the sync works is generally "last write wins", but the client will know if the same object/file has been modified on the server since last synchronizing it (by way of storing the ETAG), and in case the file has been modified both in the local and remote storage, a `conflict` event should be emitted, which apps can then use to resolve such conflicts.

When I tested this in [Webmarks](https://webmarks.5apps.com/) just now, it worked fine with new data/bookmarks, but there seems to be a bug when adding the same bookmark before connecting. After the sync, I can see the changes in the UI, but when reloading the page, or re-connecting the storage, it turns out that the changes haven't actually been synced. This can either be a bug in Webmarks or remoteStorage.js. I will investigate further to find out what is happening exactly.

### Disconnecting a remote storage

When disconnecting a remote storage from an app, the local cache is purged intentionally. It has been designed this way, because the most common use case for disconnecting a storage for us was usually wanting to connect a different account and/or to clear the local cache for other reasons.

[...]

PR #1237, opened 2021-04-02
Add option for preserving local data on disconnect

Silver-Golden:
Wanted to add a way to preserve local data on disconnect.  

There is no change to default behaviour but if preserve is set to true it does not run the _cleanups. It's functional, however I am unsure if there are negative side-effects from not running _cleanups.

refs #1170

### DIFF
--- a/src/remotestorage.ts
+++ b/src/remotestorage.ts
@@ -408,7 +408,7 @@ class RemoteStorage {
    * This method clears all stored settings and deletes the entire local
    * cache.
    */
-  disconnect (): void {
+  disconnect (preserve?:boolean): void {
     if (this.remote) {
       this.remote.configure({
         userAddress: null,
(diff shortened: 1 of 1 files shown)


raucao:
Thanks for your first pull request! I just had a look at the changes.

> however I am unsure if there are negative side-effects from not running _cleanups

Yes, this is the main task left to do here. We cannot skip all cleanups only to not remove the cached user data. You can find the `cleanup` functions in all the various modules/classes in `src` and see what they're doing.

I think one possible solution would be to set a global `preserveUserDataOnDisconnect` config, which could also be done ad-hoc when calling `remoteStorage.disconnect()` with the `preserve` argument. The respective cleanup functions could then skip deleting user data and settings on disconnect.

Another solution would be to create a function that does nothing but clear all potential caching back-ends, plus the settings in localStorage when that's available. This would have the added benefit that it could be called after having disconnected, when the user wishes to remove all local data at some point later then.

Silver-Golden:
I havent used this in years, and seems that no-one else is in favor of this feature so I am closing it out

Issue #1238, opened 2021-04-04
apiKeys in setApiKeys has the wrong TS type


Silver-Golden:
https://github.com/remotestorage/remotestorage.js/blob/8ff81f05a72ff60b1b24b00fd74c2d891e26406c/src/remotestorage.ts#L521

Currently it is as above, which is the JSDoc format I believe.  
However for typescript it could be better to repersent it as 

```typescript
interface ApiKeys {
  googledrive?: string;
  dropbox?: string;
}

setApiKeys (apiKeys: ApiKeys ): void | boolean { 
  // ...
}
```

In its current state the mismatch is causing me to use @ts-ignore 

```typescript
// current

// @ts-ignore
remoteStorage.setApiKeys({googledrive: 'key'});

// what TS expects based off the current types
remoteStorage.setApiKeys({type: 'googledrive', key: "key"});

```

raucao:
Good catch! Your solution looks good to me. Want to open a PR with that?

PR #1239, opened 2021-05-21
Fix typing for setApiKeys

eddie-dunn:
Solves: #1238


raucao (approved):
Perfect! Does exactly what it's supposed to do. :+1: 

Screenshot from 2021-05-24 11-54-39

PR #1245, opened 2021-08-07
Sync interval range is now inclusive

DougReeder:
See https://github.com/DougReeder/remoteprobe for code to test this.

### DIFF
--- a/src/remotestorage.ts
+++ b/src/remotestorage.ts
@@ -49,12 +49,12 @@ function emitUnauthorized(r) {
 }
 
 /**
-* Check if interval is valid: numeric and between 1000ms and 3600000ms
+* Check if interval is valid: numeric and between 1s and 1hr inclusive
 */
 function isValidInterval(interval: unknown): interval is number {
   return (typeof interval === 'number' &&
-          interval > 1000 &&
-          interval < 3600000);
+          interval >= 1000 &&
+          interval <= 3600000);
 }
 
 /**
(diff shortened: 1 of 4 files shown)


DougReeder:
I changed the value used in the test 'Update sync interval' - it still tests that sync interval can be set, and now we get a test of the valid range, for free.

DougReeder:
I think the range should be inclusive, to follow the "rule of least surprise".

src/remotestorage.ts
```diff
@@ -49,12 +49,12 @@ function emitUnauthorized(r) {
 }
 
 /**
-* Check if interval is valid: numeric and between 1000ms and 3600000ms
+* Check if interval is valid: numeric and between 1s and 1hr inclusive
```
raucao, line 52: I think this was intentionally more than 1 second and less than 1 hour actually. If it should be inclusive, then I think we should specify a more reasonable minimum bound here.

raucao (changes requested):
Regardless of changes proposed, please only propose one set of changes per pull request. This branch includes commits of unrelated changes, as well as merge commits. Thanks!

src/remotestorage.ts
DougReeder: 1 second *is* probably faster than anyone needs - but 1001ms isn't much better. 

My thought is that the rate is best limited by the server - one could be syncing to a private server on a private network, and be free to use a lot of resources if it made sense.  A public server should have rate-limiting, to avoid misuse and defend against DOS attacks.

DougReeder:
I'll retire this and submit two new PRs.

raucao:
Thanks!

PR #1259, opened 2022-01-18
Improve storeObject() documentation

DougReeder:
Notes that storeObject() mustn't be called more than 1/s. It's best to know this **before** writing code to save to RemoteStorage.

### DIFF
--- a/doc/js-api/base-client.rst
+++ b/doc/js-api/base-client.rst
@@ -393,7 +393,7 @@ the change event handler, nor by chaining Promises. :func:`storeObject` or
 :func:`storeFile` must not be called until the next iteration of the JavaScript
 Task Queue, using for example `setTimeout()`_.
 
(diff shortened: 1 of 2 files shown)


raucao (approved):
LGTM. 

'm still a bit uncomfortable with making this a "must" in the recommendation, and living with the fact that this is more or less an arbitrary hotfix, while there are actual improvements to be done in the sync code when queuing updates for the same path.

DougReeder:
See https://github.com/DougReeder/remoteprobe for code to test this.

DougReeder:
Read the Docs failed because "AttributeError: 'PosixPath' object has no attribute 'rstrip'".  Can someone with permission re-run it?

raucao:
I restarted the build and it failed again. Same for the latest master build.

Looks like they have updated their Docker image in a way that breaks our setup. :/

PR #1260, opened 2022-01-18
Improve validation of custom sync interval settings

DougReeder:
The "principle of least surprise" suggests that the if the upper limit is "1 hour", you should be able to use 3600000, rather than 3599999.

Minimum sync interval is now 2s and sync interval range is now inclusive.

### DIFF
--- a/src/remotestorage.ts
+++ b/src/remotestorage.ts
@@ -49,12 +49,12 @@ function emitUnauthorized(r) {
 }
 
 /**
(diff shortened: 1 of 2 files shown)


raucao (approved):
:+1:

galfert:
This should be noted as a breaking change in the changelog. Apps using a sync interval of e.g. 1500 will get an error thrown now.

PR #1266, opened 2022-03-23
Add more types, improve syntax and formatting

raucao:


raucao:
FYI: I have tested this branch with both an app of mine as well as the TypeScript example repo.

PR #1269, opened 2022-05-27
Throw if connect() is called before claiming access

### LINKED ISSUE #1198: Check if access is claimed when connecting
I just added the library to a test app and forgot to claim a scope. It will just fail with an error message, but not a helpful one for new developers.

ssisk:
Fixes #1198

Hello! I ran into this issue and I thought I'd go ahead and submit a fix since it seems like a nice small change.

### DIFF
--- a/test/unit/authorize-suite.js
+++ b/test/unit/authorize-suite.js
@@ -120,12 +120,12 @@ define([ 'require', './build/authorize', './build/unauthorized-error'], function
 
           var sawError = false;
           try {
-          env.Authorize.authorize(this, {authURL, scope, redirectUri, clientId});
+            env.Authorize.authorize(this, {authURL, scope, redirectUri, clientId});
           } catch (e) {
-            test.assert(e.message.includes("scope"), true, "Did not throw an error that had a message describing the problem");
+            test.assert(e.message.includes("scope"), true, "Thrown Error must have a message mentioning 'scope'");
             sawError = true;
           }
-          test.assert(sawError, true, "Did not throw an error when scope was not set");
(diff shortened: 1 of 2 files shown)


DougReeder (approved):

test/unit/authorize-suite.js
```diff
@@ -108,6 +108,27 @@ define([ 'require', './build/authorize', './build/unauthorized-error'], function
...
+          this.localStorageAvailable = function() { return true; };
+
+          var sawError = false;
+          try {
+          env.Authorize.authorize(this, {authURL, scope, redirectUri, clientId});
+          } catch (e) {
+            test.assert(e.message.includes("scope"), true, "Did not throw an error that had a message describing the problem");
+            sawError = true;
+          }
+          test.assert(sawError, true, "Did not throw an error when scope was not set");
```
raucao, line 128: The descriptions of the assertions should be human-readable language of what is being asserted. When the test fails, it will then show that e.g. "throws an error when scope is not set" was expected, but did not satisfy the expectation.

raucao (changes requested):
Welcome @ssisk! And thanks for your first contribution! :tada: 

I just have a minor nitpick about wording, but overall it's looking good to me.

test/unit/authorize-suite.js
ssisk: Thanks for the feedback! I've updated the assertions' text so that they should match the convention, let me know if not.

raucao (approved):
:+1:

test/unit/authorize-suite.js
raucao: Great, thanks!

PR #1276, opened 2022-11-03
Fix ReadTheDocs builds

### LINKED ISSUE #1277: ReadTheDocs builds broken (again)
Unfortunately, commit 3af7ff92 broke the documentation builds by upgrading TypeDoc without ensuring that the builds still work. This has slipped through review before the PR was merged, so it also broke the builds in master since then.

raucao:
Unfortunately, there were a lot of breaking changes since `typedoc` 0.19.2, and SphinxJS doesn't work with those yet. However, the usage of `XMLHttpRequestBodyInit` since recent refactors requires a newer version of TypeScript than 0.19.2 specifies as allowed peer dependency.

NodeJS 14 doesn't complain about the mismatch when mixing old `typedoc` with new `typescript`, but NodeJS 16 does. So with this fix, it is necessary to use either the `--force` or `--legacy-peer-deps` flag with `npm install` and `npm ci` for now.

fixes #1277

### DIFF
--- a/.github/workflows/test-and-lint.yml
+++ b/.github/workflows/test-and-lint.yml
@@ -20,7 +20,7 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
       - name: Install dependencies
-        run: npm ci
+        run: npm ci --force
       - name: Run jaribu tests   # These must be replaced before we can use Node v18 in CI
         run: npm test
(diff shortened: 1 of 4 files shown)


galfert (approved):
Looks ok to me. Sometimes things just need to be forced

sandwich

PR #1282, opened 2022-12-13
Fix getNodes does not clone object in inmemorystorage

### LINKED ISSUE #1281: getNodes does not clone object in inmemorystorage
Ok, thanks. Confirmed.

> On a side note, should I open a bug for the fix I applied?

Yes, I think that's a good idea.

_Originally posted by @raucao in https://github.com/remotestorage/remotestorage.js/issues/1279#issuecomment-1346528515_

darkdread:
Getting a node in memory storage should clone whatever it was holding to prevent data mutation.

fixes #1281

### DIFF
--- a/src/inmemorystorage.ts
+++ b/src/inmemorystorage.ts
@@ -1,7 +1,7 @@
 import type { RSNode, RSNodes } from './interfaces/rs_node';
 import EventHandling from './eventhandling';
 import CachingLayer from './cachinglayer';
-import { applyMixins } from './util';
+import { applyMixins, deepClone } from './util';
 
 /**
  * In-memory caching adapter. Used when no IndexedDB or localStorage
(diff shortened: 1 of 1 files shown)


raucao (commented):
Just FYI, the localStorage back-end also doesn't seem to do a clone there...

raucao (approved):

darkdread:
> Just FYI, the localStorage back-end also doesn't seem to do a clone there...

https://github.com/remotestorage/remotestorage.js/blob/a189529ec1b589094131d708c6442da0fad72c04/src/localstorage.ts#L39

https://github.com/remotestorage/remotestorage.js/blob/a189529ec1b589094131d708c6442da0fad72c04/src/localstorage.ts#L51

Nodes are objects, and objects are mutable in JavaScript.

When we store nodes in localStorage, we stringify the object. When we get nodes from localStorage, we parse the string back into an object. This makes the node immutable, which means modifying the node outside of `getNodes` will not mutate localStorage, because the object reference is not referencing the node in localStorage.

What we have initially in inmemorystorage however, was referencing the node stored in `this_storage`, which means modifying any node outside of `getNodes` will modify the node in `this_storage`.

PR #1295, opened 2023-03-30
Lint mocha specs in CI

raucao:
Let's ensure somewhat consistent syntax in the new mocha spec files. Using the same rules as source code, but overriding a few to make more sense with specs (e.g. turning off maximum statements per function).


DougReeder (approved):
LGTM

PR #1296, opened 2023-04-10
Update git-up recommendations with git-native commands

raucao:
The recommended solution is not needed anymore, since the same can now be done with native Git options and commands.


DougReeder (approved):
LGTM

PR #1298, opened 2023-07-13
Add note about RS Inspektor

Nezteb:
Related to https://github.com/remotestorage/remotestorage-server/issues/6 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ remoteStorage server implementation: [Servers](https://wiki.remotestorage.io/Ser
 
 ### Visual File Browser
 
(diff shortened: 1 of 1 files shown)


DougReeder (approved):

README.md
```diff
@@ -42,6 +42,10 @@ local test server, or for self-hosting an RS server.
 You can also get an account with a hoster, or use another
 remoteStorage server implementation: [Servers](https://wiki.remotestorage.io/Servers).
 
+### Visual File Browser
+
+If you'd like a visual UI for inspecting any RS-compatible server, you can try out [RS Inspektor](https://inspektor.5apps.com) or [build it from source](https://gitea.kosmos.org/raucao/inspektor), Feel free to ask questions or propose features for this app on the [RS forums](https://community.remotestorage.io)!
```
raucao, line 47: I have updated Inspektor's README to link to the forums, so we could shorten it a little bit in the README here. How about this?

```suggestion
If you'd like a visual UI for inspecting any RS-compatible account, you can use the [RS Inspektor](https://gitea.kosmos.org/raucao/inspektor) app (which is also implemented using this library).
```

raucao (approved):

Issue #1270, opened 2022-09-06
Tests must be migrated away from Jaribu, which is no longer maintained


DougReeder:
https://github.com/silverbucket/jaribu/blob/master/README.md

raucao:
Anyone interested in contributing, see #1193, where I started implementing this a while back.

silverbucket:
I would like to help out, just don't know if I realistically have the time right now. What about getting #1193 merged so we can pick at it as we go? 

Any new tests could be implemented with the new framework, and as old tests need to be refactored they can be rewritten as well. That's how I've been doing it with Sockethub, and even though there are still a few remaining Jaribu tests, most of the codebase in covered with mocha and chai now.

PR #1299, opened 2023-10-28
Security updates from npm audit fix; shouldn't be any breaking changes

DougReeder:
If all packages here are using semantic versioning properly, there shouldn't be any breaking changes. Do we need any testing beyond the existing testing suite?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -59,12 +59,13 @@
       }
     },
     "node_modules/@babel/code-frame": {
-      "version": "7.18.6",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.18.6.tgz",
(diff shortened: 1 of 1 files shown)


raucao:
> Do we need any testing beyond the existing testing suite?

The test suite should catch any issues, but a quick smoke test in a real app never hurts.

The node 16 suite is currently failing for this PR. And I guess we should add current LTS (20.0) and perhaps 18 (still widely used) to the matrix.

DougReeder:
Smoke test completed, using Notes Together and nodeJS v16.20.2 on MacOS 14.0.

The earlier failure of the test suite was a timeout, which did not happen on a re-run.

As for nodeJS v18, both the Jaribu and Mocha tests fail with "ReferenceError: FileReader is not defined", so that will have to wait. Also note [Jaribu is no longer supported](https://github.com/remotestorage/remotestorage.js/issues/1270)

raucao (approved):

Issue #1300, opened 2023-11-03
Tests fail under NodeJS v18


DougReeder:
Both the  `test` and `test:mocha` scripts fail under node 18 with the error: `ReferenceError: FileReader is not defined`

changing 
`if (typeof Blob === 'undefined') {`
to    `if (typeof Blob === 'undefined' || typeof FileReader === 'undefined') {`
at line 248 of `util.ts` fixes this for `test:mocha`, but then `test` throws 
```
UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "[object Object]".
```

raucao:
I had a look at this, and it's actually just due to `Blob` having promoted to the global context in node 18, since it's not experimental anymore, and our code checks for Blob being available or not to determine if it should read the file using browser APIs or node.js ones. Fix incoming.

PR #1307, opened 2024-01-16
Update Node.js versions in CI

raucao:


raucao:
Sounds good to me. Want to have go at that?

DougReeder:
The error message is "test failed ReferenceError: FileReader is not defined". IIRC, it has something to do with what APIs are available in which version of Node.

raucao:
@silverbucket See here: https://github.com/remotestorage/remotestorage.js/issues/1300

raucao:
The remaining test failures are not due to the `FileReader` issue. Only the Webfinger tests are failing, while the `WireClient` ones are passing. PR with actual issue description incoming...

PR #1313, opened 2024-06-16
Support node.js 18, 20

### LINKED ISSUE #1300: Tests fail under NodeJS v18

### LINKED ISSUE #1308: Jaribu test failing in CI for node 16
It started failing a while ago, but there is no obvious change that I could find, which would explain the exception. Furthermore, I have not been able to reproduce the failure on my machine so far.

raucao:
See commits for details.

fixes #1300, fixes #1308, refs #1270


raucao:
The next step is fixing the documentation dependency hell, which also requires us to use `--force` for npm installs currently. I think I'll have a go at converting everything to Markdown, and then trying to get as close as possible to the current docs using VitePress and https://www.typedoc-plugin-markdown.org/plugins/vitepress, so we can ditch the whole Python/Sphinx setup entirely and use the latest versions of typedoc and typescript.

DougReeder (approved):
These are all good changes to modernize the code.

PR #1324, opened 2025-05-30
Remove all ReadTheDocs remnants

raucao:
We missed a few things. See commit.


DougReeder (approved):
LGTM and new URLs are correct

PR #1325, opened 2025-05-30
Automate website updates

raucao:
This adds an Action to notify the website repo of updates to this repo's master branch, which should trigger a website build and deployment.

I haven't tested this yet, but in theory it should trigger this action that I pushed to the website repo already:

https://github.com/remotestorage/website/blob/master/.github/workflows/update-rsjs-submodule.yml

Maybe someone could have a quick look to spot any blunders, before we try this in production.

(There's also a change for building and committing the typedoc API docs upon release via `npm version`.)

### DIFF
--- a/.github/workflows/notify-website-repo.yml
+++ b/.github/workflows/notify-website-repo.yml
@@ -1,0 +1,17 @@
+name: Notify website repo
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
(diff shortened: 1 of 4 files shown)


DougReeder (approved):
I haven't worked with these mechanisms before, but it looks okay.

raucao:
OK, let's try.

PR #1327, opened 2025-06-03
Improve connect docs

raucao:
The current doc only describes using the widget, but it's missing the actual connect function entirely.

This documents both options, and also adds another tip for a hybrid widget/no-widget approach.


DougReeder (approved):
LGTM

Issue #1251, opened 2021-10-18
Refactor sync/caching


raucao:
This was one of the main reasons for porting the source code to TypeScript. So now it's time to actually refactor and clean up the code that is virtually impossible to quickly debug due to extended Promise chains, some of which span 20+ functions. (Just FYI, for new contributors: when that code was written, Promise wasn't even part of JavaScript itself, but only a half-standardized spec which we used via a utility library).

My proposal would be to use async/await as much as possible instead.

Edit: I have added sub issues for everything I could find that is related and would be good to close on the way.

PR #1331, opened 2025-06-08
Type, document, and refactor Sync, port tests to Mocha

### LINKED ISSUE #839: Suppress SyncErrors when in offline mode
When I shutdown my connection, I get the following errors:

``` javascript
[Sync] Error SyncError: Sync failed: Network request failed. {originalError: "Network request failed.", message: "Sync failed: Network request failed.", constructor: function}
```

``` javascript
remotestorage.min.js:4 OPTIONS https://storage.5apps.com/leblowl/collections/ net::ERR_INTERNET_DISCONNECTED
```

For me these are ok, I just expect them to be log messages or warnings rather than exceptions since "offline" is an accepted state. I like to keep my console clean when not using debug mode unless a fatal error has occurred. So is there any built-in way to reduce their error level? thank you!

### LINKED ISSUE #1173: Add `sync-started` event
Currently, the widget and app authors have to rely on the first `sync-req-done` to know that a sync is in progress. I think it would be much better to have a dedicated `sync-started` event, which would be fired immediately, when a sync has started.

### LINKED ISSUE #1204: IndexedDB entries never delete in anonymous mode
It seems like deleting objects in anonymous mode marks them as `local.body: false` instead of removing their entries from the database. I tested in Safari (non-private) and Chrome (private) using my test app locally as well as Webmarks. I can clear them using the browser developer tools but not with the library. In Chrome I tried connecting an account and it seems to delete the entries as expected.

### LINKED ISSUE #1335: Empty documents not handled correctly, can lead to failed PUTs
This line in the [codebase](https://github.com/remotestorage/remotestorage.js/blob/a4bebfa84c4e99675a450c1281e9edbb546a41e3/src/sync.ts#L466) causes issue if the body is 0 byte; it will skip the storing of revision (etag) and subsequent pushes to the server will fail due to etag mismatch.

The correct behavior I can think of are:

1) Keep the current behavior, but the remotestorage client MUST store 0 byte document properly
2) Accept 0 byte document, but allow PUT with/without etag (treat as empty)
3) Disallow 0 byte document at all

Maybe this issue should be referenced in https://github.com/remotestorage/spec/issues?

raucao:
Includes some small other improvements. However, this PR is mostly about properly typing and documenting sync, as well as making the functions more readable by using async/await in many places.

Noteworthy additional changes:

* Adds a new event called "sync-started", when a periodic/background sync has started (but not when outgoing changes are synced immediately).
* Emit conflict change events asynchronously, so app developers don't have to handle them in a special way, improve conflict event docs
* Do not emit conflict events for nodes that were deleted both locally and remotely (d04f334)
* Fix remote deletes not handled properly when folder has local changes (098962db)
* For more, see referenced issues below

refs #1251 (where I added sub-issues that are all easier to tackle with these changes)
refs #1270
closes #1173
closes #839
fixes #1335
fixes #1204

P.S.: I rebased the branch as much as possible before opening the PR. But there are too many conflicts when trying to consolidate the rest even more.

### DIFF
--- a/src/sync.ts
+++ b/src/sync.ts
@@ -729,7 +729,7 @@ export class Sync {
         collectSubPaths(node.common, path);
         collectSubPaths(node.local, path);
       } else {
-        if (node.common && typeof(node.common.body) !== undefined) {
+        if (node.common && typeof(node.common.body) !== 'undefined') {
           changedNodes[path] = deepClone(node);
           changedNodes[path].remote = {
             body:      false,


raucao:
FYI, I opened #1334 against this one, because it requires the new cachinglayer test suite.

raucao:
We'll have to merge this soon. It's been too long, and I cannot add anymore PRs against this one.

rosano:
> Bonus: if your app uses change events with origin `window`, it should even work between browser tabs in the same browser session now (and instantly).

Looking forward to implementing that in future apps, and maybe documenting.

src/sync.ts
```diff
@@ -554,134 +622,136 @@ export class Sync {
...
-              body:      false,
-              timestamp: this.now()
-            };
-            changedNodes[path] = this.autoMerge(changedNodes[path]);
-          }
+      if (isFolder(path)) {
+        collectSubPaths(node.common, path);
+        collectSubPaths(node.local, path);
+      } else {
+        if (node.common && typeof(node.common.body) !== undefined) {
```
galfert, line 732: `typeof` always returns a string.

```suggestion
        if (node.common && typeof(node.common.body) !== 'undefined') {
```

galfert (approved):

PR #1332, opened 2025-06-09
Add debug logging for visibility changes

raucao:
Nothing much to say. Debug logs are supposed to be verbose, so I think visibility mode changes should appear there as well.


galfert (approved):
Makes sense 👍

PR #1334, opened 2025-06-09
Broadcast change events to other browser tabs

### LINKED ISSUE #566: events with 'device' origin don't get emitted
when having the same app open in two windows or tabs open in the same browser on the same device, you will miss change events for changes that are made from other tabs.

we used to use StorageEvents for this in localStorage i think back in the 0.7 days, but we'll have to do something else in indexeddb; this part was simply was never implemented in the 0.8 rewrite.

### LINKED ISSUE #1137: Change event issues with tabs/window
While thoroughly testing an unrelated pull request, I found a problem with our change events. They used to work as intended, but at the moment, with latest Firefox and Chrome, it seems that they are slightly broken, when syncing the same category across multiple tabs in the same window.

The intended behaviour is to receive a `window` change event in the other tabs, and to always receive `remote` change events in all tabs (although I'm not certain about that latter).

The behaviour I'm seeing when changing an object is:

1. Window event in the same tab being emitted, but not received by my app.
2. No event emitted or received in the other tab. Both window and remote missing.
3. When sent from a different browser/window, only one of the two tabs receives the remote change event.

raucao:
Simple, but amazing UX. You can't change tabs fast enough to see the changes being applied live. :tada: 

closes #1137
closes #566

### DIFF
--- a/src/cachinglayer.ts
+++ b/src/cachinglayer.ts
@@ -2,6 +2,7 @@ import type { ChangeObj } from './interfaces/change_obj';
 import type { QueuedRequestResponse } from './interfaces/queued_request_response';
 import type { RSEvent } from './interfaces/rs_event';
 import type { RSItem, RSNode, RSNodes, ProcessNodes } from './interfaces/rs_node';
+import Env from './env';
 import EventHandling from './eventhandling';
 import config from './config';
 import log from './log';
@@ -9,6 +10,7 @@ import {
   applyMixins,
   deepClone,
   equal,
+  globalContext,
   isDocument,
   isFolder,
   pathsFromRoot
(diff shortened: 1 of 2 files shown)


src/cachinglayer.ts
```diff
@@ -120,6 +122,21 @@ abstract class CachingLayer {
...
   abstract setNodes(nodes: RSNodes): Promise<void>;
 
+  /**
+   * Broadcast channel, used to inform other tabs about change events
+   */
+  broadcastChannel: BroadcastChannel;
+
+  constructor () {
+    const env = new Env();
+    if (env.isBrowser() && !!globalContext["BroadcastChannel"]) {
```
DougReeder, line 132: https://developer.mozilla.org/en-US/docs/Web/API/BroadcastChannel notes that BroadcastChannel is available in Node.js since v18 and Deno since 1.11.  Do we need to confine this to browsers?

PR #1356, opened 2026-03-15
Fix sync skipping PUT for ArrayBuffer bodies

### LINKED ISSUE #1354: Sync silently skips PUT for ArrayBuffer bodies stored via storeFile
`storeFile()` accepts `string | ArrayBuffer | ArrayBufferView` per its type signature, but the sync layer only pushes `string` bodies. ArrayBuffer bodies are stored in IndexedDB but never synced to the remote server — silently, with no error.

### Steps to reproduce

1. Connect to a remoteStorage server (tested with Armadietto)
2. Enable caching: `rs.caching.enable('/test/')`
3. Store a file with an ArrayBuffer body:
   ```js
   const buf = new Uint8Array([72, 101, 108, 108, 111]).buffer;
   await client.storeFile('application/octet-stream', 'test.bin', buf);
   ```
4. Wait for `sync-done` event
5. Check the server — the file was never uploaded

### Expected behavior

The file should be PUT to the remote server during sync.

### Actual behavior

- `storeFile` resolves successfully (file is in IndexedDB)
- Sync runs and completes — no PUT request is made
- `sync-done` fires with `{completed: true}`
- The file exists only in IndexedDB and is lost if the user clears app data

### Root cause

`needsRemotePut()` in sync.ts:

```typescript
needsRemotePut(node) {
  return node.local && typeof(node.local.body) === "string";
}
```

`typeof ArrayBuffer` is `"object"`, so this returns `false`. The sync task falls through to a GET instead of a PUT. No error is thrown.

Meanwhile, `storeFile` in baseclient.ts explicitly accepts ArrayBuffer:

```typescript
async storeFile(contentType: string, path: string, body: string | ArrayBuffer | ArrayBufferView)
```

This is an internal inconsistency — the public API accepts ArrayBuffer, but the sync layer silently ignores it.

### Possible fix

Either:
- `needsRemotePut` should check for `node.local.body !== undefined && node.local.body !== false` (or similar) instead of restricting to strings
- Or `storeFile` / `storage.put` should convert ArrayBuffer to string before storing locally

### Environment

- remotestoragejs 2.0.0-beta.8
- Server: Armadietto
- Browsers: Chrome, Firefox

silverbucket:
## Summary
- `needsRemotePut()` checked `typeof(body) === "string"`, silently skipping ArrayBuffer/ArrayBufferView bodies that `storeFile()` accepts
- Changed the check to `body !== undefined && body !== false` so all valid body types trigger a remote PUT during sync
- Updated the type cast in `doTask` to match the actual body types

Closes #1354

## Test plan
- [x] Added unit tests for `needsRemotePut` covering string, empty string, ArrayBuffer, Uint8Array, `false` (delete), and undefined local
- [x] All 96 sync tests pass

### DIFF
--- a/docs/api/baseclient/classes/BaseClient.md
+++ b/docs/api/baseclient/classes/BaseClient.md
@@ -723,9 +723,6 @@ A new `BaseClient` operating on a subpath of the current base path
 
 Store raw data at a given path.
 
-For binary data, use an `ArrayBuffer` or `ArrayBufferView`
-(e.g. `Uint8Array`), not a binary string.
-
 #### Parameters
 
 • **contentType**: `string`
--- a/src/baseclient.ts
+++ b/src/baseclient.ts
@@ -466,12 +466,10 @@ export class BaseClient {
   /**
    * Store raw data at a given path.
    *
-   * For binary data, use an `ArrayBuffer` or `ArrayBufferView`
-   * (e.g. `Uint8Array`), not a binary string.
-   *
    * @param contentType - Content type (MIME media type) of the data being stored
    * @param path        - Path relative to the module root
-   * @param body        - Raw data to store
+   * @param body        - Raw data to store. For binary data, use an `ArrayBuffer`
+   *                      or `ArrayBufferView` (e.g. `Uint8Array`), not a binary string.
    *
    * @returns A promise for the created/updated revision (ETag)
    *


test/unit/sync.test.mjs
```diff
@@ -805,6 +805,38 @@ describe("Sync", function() {
...
+
+      it("returns true for an ArrayBuffer body", function() {
+        const node = { local: { body: new ArrayBuffer(8), contentType: "application/octet-stream" } };
+        expect(this.rs.sync.needsRemotePut(node)).to.be.true;
+      });
+
+      it("returns true for a Uint8Array body", function() {
+        const node = { local: { body: new Uint8Array([1, 2, 3]), contentType: "application/octet-stream" } };
+        expect(this.rs.sync.needsRemotePut(node)).to.be.true;
+      });
```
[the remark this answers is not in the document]
silverbucket: Added `doTask` integration tests that stub `remote.put` and verify both `ArrayBuffer` and `Uint8Array` bodies are passed intact with correct bytes.

src/sync.ts
```diff
@@ -425,7 +425,7 @@ export class Sync {
           return taskFor('put', path,
             this.rs.remote.put(
               path,
-              node.push.body as string, // TODO string | ArrayBuffer?
+              node.push.body as string | ArrayBuffer | ArrayBufferView,
               node.push.contentType,
```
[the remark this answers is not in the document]
silverbucket: Valid suggestion but out of scope for this fix — updating `RSItem.body` in `rs_node.ts` would touch type definitions across the codebase (cachinglayer, indexeddb, sync, wireclient). Leaving for a separate PR.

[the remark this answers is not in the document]
silverbucket: Addressed. Updated `RSItem.body` and `previousBody` in `rs_node.ts` to include `ArrayBuffer | ArrayBufferView`. Also updated `CachingLayer.put` signature, `corruptRevision` validation, and simplified the cast in `doTask` to `XMLHttpRequestBodyInit`. All type-safe end-to-end now.

src/util.ts
```diff
@@ -18,6 +18,15 @@ function _fixArrayBuffers(srcObj: object, dstObj: object) {
         const srcArr = new Int8Array(srcObj[field]);
         const dstArr = new Int8Array(dstObj[field]);
         dstArr.set(srcArr);
+      } else if (ArrayBuffer.isView(srcObj[field])) {
```
DougReeder: This works fine, but can we just replace `_deepClone` with `structuredClone` which is available on all modern browsers and all maintained versions of Node.js, and explicitly handles ArrayBuffers and DataViews (and, really, anything that makes sense)?   https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#supported_types

src/sync.ts
```diff
@@ -229,7 +229,7 @@ export class Sync {
     return ((typeof(rev) !== 'object') ||
             (Array.isArray(rev)) ||
             (rev.revision && typeof(rev.revision) !== 'string') ||
-            (rev.body && typeof(rev.body) !== 'string' && typeof(rev.body) !== 'object') ||
+            (rev.body && typeof(rev.body) !== 'string' && typeof(rev.body) !== 'object' && !(rev.body instanceof ArrayBuffer) && !ArrayBuffer.isView(rev.body)) ||
```
raucao: `typeof new ArrayBuffer(0)` evaluates to `object`, so these additions are a dead code path. I think we can just remove them:

```suggestion
            (rev.body && typeof(rev.body) !== 'string' && typeof(rev.body) !== 'object')  ||
```

src/util.ts
```diff
@@ -178,9 +153,7 @@ export const deepClone = (obj: any): any => {
   if (obj === undefined) {
     return undefined;
   } else {
-    const clone = JSON.parse(JSON.stringify(obj));
-    _fixArrayBuffers(obj, clone);
-    return clone;
+    return structuredClone(obj);
```
raucao, line 156: This breaks backward compatibility with anything older than 2022, which will include a lot of Android web views on older phones and such. How about we make it optional, with something like this e.g.:

```ts
export const deepClone = (obj: any): any => {
  if (obj === undefined) {
    return undefined;
  } else if (typeof structuredClone === 'function') {
    // Fast path for modern browsers (2022+) and Node.js 17+
    return structuredClone(obj);
  } else {
    // Fallback for older browsers
    const clone = JSON.parse(JSON.stringify(obj));
    _fixArrayBuffers(obj, clone);
    return clone;
  }
};
```

silverbucket: The fallback means keeping _fixArrayBuffers around, which is the whole thing this commit was trying to eliminate. 

...And I *think* it means that we still can’t handle typed arrays properly (the original bug this PR fixes), so we would have two code paths where one silently corrupts binary data on older browsers anyway.

Maybe something like a polyfill during build?

silverbucket: What about:

```javascript
export const deepClone = (obj: any): any => {
  if (obj === undefined) {
    return undefined;
  } else if (typeof structuredClone === 'function') {
    return structuredClone(obj);
  } else {
    return JSON.parse(JSON.stringify(obj));
  }
};
```

This drops _fixArrayBuffers entirely. On old browsers the JSON fallback would still work for strings and objects (the only body types the old code supported anyway). Binary bodies were already broken on those browsers before this PR.

raucao: Yes, let's get rid of it. :+1: 

I think we should also change the documentation to prescribe not using strings at all for binary data and document it as breaking change.

Btw, it was only broken with caching enabled, not for clients without (e.g. Sharesome), right?

silverbucket:
@raucao addressed your points, this is ready for your final approval and merge

docs/api/baseclient/classes/BaseClient.md
```diff
@@ -723,6 +723,12 @@ A new `BaseClient` operating on a subpath of the current base path
 
 Store raw data at a given path.
 
+For binary data, always use an `ArrayBuffer` or `ArrayBufferView`
+(e.g. `Uint8Array`) — not a binary string. String bodies are only
+reliably synced on environments that support `structuredClone`; on
+older browsers without it, binary strings will lose fidelity during
+the internal deep-clone step.
+
```
raucao: Shouldn't this go in the parameter description below?

(Unfortunately, I cannot collaborate on this PR's code, because it was open in your personal fork.)

raucao: Oh, actually this was generated but shouldn't be checked in. Docs are updated automatically on release...

silverbucket:
@raucao OK, resolved the doc issues. Ready again for approval

raucao:
bluesbrothers

PR #1362, opened 2026-04-12
Add LLM/agent configs

raucao:
Add AGENTS.md with general instructions, and an OpenCode config that permits only necessary commands to be run

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,188 @@
+# Scope
+
+- This file applies to the entire repository: `remotestorage/remotestorage.js`.
+- Follow these instructions for all code changes, scripts, and tests.
+
+# Build, Lint, Test
+
+- Install dependencies: `npm ci` (CI) or `npm install` (local).
+- TypeScript compile: `tsc` or `tsc -w` for auto-compile on changes.
+- Development bundle/watch: `npm run dev`.
+- Production bundle: `npm run build:release` (webpack production).
+- Lint sources: `npm run lint` (verbose) or `npm run lint:quiet`.
+- Lint Mocha specs: `npm run lint:specs` or `npm run lint:specs:quiet`.
+- Full test suite: `npm test` (runs `tsc` and `scripts/test-all.sh`).
+- Mocha unit tests: `npm run test:mocha`.
+- Mocha watch: `npm run test:watch`.
+- Typedoc docs (CI release step): `typedoc` via `npm run version`.
+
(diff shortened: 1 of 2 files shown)


AGENTS.md
```diff
@@ -0,0 +1,188 @@
...
+  - Single test by grep: `npm run test:mocha -- --grep "pattern"`.
+  - Watch a file: `npm run test:watch -- test/unit/<name>.test.mjs`.
+- Jaribu (legacy suites while being ported):
+  - Single suite: `./node_modules/.bin/jaribu test/unit/<suite>-suite.js`.
+  - Notes: Jaribu suites are older `.js` files like `test/unit/inmemorycaching-suite.js`.
+
+# Continuous Integration
+
+- GitHub Actions workflow: `.github/workflows/test-and-lint.yml`.
+- Matrix Node versions: `18`, `20`.
```
DougReeder, line 32: One of the many reasons to migrate away from Jaribu, is some tests wouldn't run under recent versions of Node.js.  However, we appear to have migrated the tests that specifically broke under Node.js, so we should probably be testing under Node.js v22 and v24 now

DougReeder (commented):
This is certainly useful as it stands.

According to the [GitHub Blog](https://github.blog/ai-and-ml/github-copilot/how-to-write-a-great-agents-md-lessons-from-over-2500-repositories/) we should write different versions of AGENTS.md for different roles. I can't say from my own experience which way is better in practice.

AGENTS.md
raucao: True. I agree that we should always support at least the previous and current LTS. These instructions only reflect the current code/CI. PRs welcome!

raucao:
> According to the [GitHub Blog](https://github.blog/ai-and-ml/github-copilot/how-to-write-a-great-agents-md-lessons-from-over-2500-repositories/) we should write different versions of AGENTS.md for different roles. I can't say from my own experience which way is better in practice.

Sorry, I shouldn't have merged this without waiting for reviews. It was the result of me cleaning up my local Git history across various repos a bit too hastily.

I don't actually have any experience with agents that aren't just me clearly and very specifically stating my intents to the LLM in something like OpenCode and treating the LLM like a junior pair programmer. From that experience, I'm currently not overly interested in async/remote agents doing all kinds of things, because I don't want to be a nanny to a whole bunch of robots doing things I'm not overseeing directly.

That said, feedback and PRs from people with more experience are more than welcome!

PR #1365, opened 2026-04-15
Update dependencies

raucao:


raucao:
@silverbucket This issue is blocking the webfinger.js upgrade: https://github.com/silverbucket/webfinger.js/issues/150

silverbucket:
> @silverbucket This issue is blocking the webfinger.js upgrade: [silverbucket/webfinger.js#150](https://github.com/silverbucket/webfinger.js/issues/150)

Can you give 3.0.2 a try? Just published

raucao:
Hmm, it's not on npm (yet?)

https://github.com/silverbucket/webfinger.js/issues/153

silverbucket:
> Hmm, it's not on npm (yet?)

Had an issue with the token, it's published now. 
https://www.npmjs.com/package/webfinger.js/v/3.0.2

raucao:
Found another issue: https://github.com/silverbucket/webfinger.js/issues/155

raucao:
Since the Webfinger.js API has changed, the changes necessary to upgrade to 3.x (once the properties bug has been fixed) are too much for this PR. So let's merge this one first.

The good news is that I have found the mysterious dangling requests that also held up the finishing of the test suite from time to time: turns out they were Webfinger.js fallback requests.

tsconfig.json
```diff
@@ -2,11 +2,15 @@
   "compilerOptions": {
     "target": "es2015",
```
raucao, line 3: By the way, I think we should change the target to `es2017`, which natively includes async/await.

silverbucket:
> once the properties bug has been fixed

Published the fix 3.0.3

raucao:
@silverbucket Saw it, thanks! Can you review/approve this PR?

(The target change should also be its own PR, because it's a breaking change that isn't updating dependencies.)

silverbucket (approved):

PR #1371, opened 2026-04-17
Fix occasional unmatched requests in Dropbox suite

raucao:
Prevents unwanted background sync in Dropbox suite which may result in the suite completely successfully, but not hanging on exit (due to an unresolved auth request).


DougReeder (approved):
The tests pass, and this is reasonable.

PR #1372, opened 2026-04-19
Updates CI to test under Node v22 & v24

DougReeder:
Why: so we're warned of incompatibilities while we have plenty of time to fix them

### DIFF
--- a/.github/workflows/test-and-lint.yml
+++ b/.github/workflows/test-and-lint.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       matrix:
         # Support LTS versions based on https://nodejs.org/en/about/releases/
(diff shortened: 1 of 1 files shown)


.github/workflows/test-and-lint.yml
```diff
@@ -14,16 +14,17 @@ jobs:
     strategy:
       matrix:
         # Support LTS versions based on https://nodejs.org/en/about/releases/
-        node-version: ['18', '20']
+        node-version: ['18', '20', '22', '24']
```
raucao, line 17:
```suggestion
        node-version: ['20', '22', '24']
```

[18 is officially EOL](https://nodejs.org/en/about/previous-releases), so I think we should remove it.

raucao (approved):

Issue #1180, opened 2020-04-11
Hash is removed if it contains an `=`.


kevincox:
Right now if the hash contains an `=` it is assumed to "belong" to the remotestorage.js oauth flow and is removed in its entirely. This is very annoying for websites that want to use the hash themselves. Instead the hash should only be removed if there is strong evidence to indicate that we are doing the remotestorage.js oauth flow (for example `rsDiscovery` or `access_token` and `state` are set). Ideally even stronger verification could be done such as embedding a random value that is unlikey to be used by the embedding application.

https://github.com/remotestorage/remotestorage.js/blob/61cbbb25a99e1ce995ddf0fac8571683c32e9b8b/src/authorize.js#L215-L220

silverbucket:
resolved in #1374

PR #1374, opened 2026-04-24
Fix authorize hash cleanup for non-OAuth fragments

### LINKED ISSUE #1373: `location.hash` is cleared on every construction, not just OAuth callbacks
## Problem

`new RemoteStorage(...)` unconditionally wipes `window.location.hash` on every page load — not just after OAuth callbacks. This breaks hash-based SPA routers (`#/route`) that use the fragment for navigation: any route hash (e.g. `#/todos`, `#/collections`) is stripped synchronously during construction, and the resulting `hashchange` sends the router to its default route.

## Reproduce

```js
// URL: https://example.com/#/todos
import RemoteStorage from 'remotestoragejs';
const rs = new RemoteStorage({});
// URL after construction: https://example.com/#
// (hashchange fires: #/todos → #)
```

## Root cause

[`src/authorize.ts` lines 271-278](https://github.com/remotestorage/remotestorage.js/blob/master/src/authorize.ts#L271-L278):

```ts
static _rs_init = function (remoteStorage: RemoteStorage): void {
  const params = extractParams();
  let location: Location;

  if (params) {
    location = Authorize.getLocation();
    location.hash = '';
  }
  ...
```

The `if (params)` guard is intended to mean "only clear when this is an OAuth callback", but `extractParams()` always returns at least `{}` (empty object), and `{}` is truthy in JavaScript. So the guard never fails and the hash is always cleared.

## Scenarios

| URL hash | Query string | `extractParams()` returns | `if (params)` | Hash cleared? |
|---|---|---|---|---|
| `#/todos` | none | `{}` | truthy | ✓ (bug) |
| `#foo=bar` | none | `{foo: 'bar'}` | truthy | ✓ (arguably bug — see #1180) |
| (empty) | `?q=1` | `{q: '1'}` | truthy | ✓ (no-op) |
| (empty) | none | `{}` | truthy | ✓ (no-op) |

## Suggested fix

Gate the cleanup on an actual OAuth-callback shape rather than a truthy-object check:

```ts
const isOAuthCallback = !!(
  params.access_token ||
  params.code ||
  params.error ||
  params.remotestorage ||
  params.rsDiscovery ||
  params.state
);

if (isOAuthCallback) {
  location = Authorize.getLocation();
  location.hash = '';
}
```

This aligns with the intent of #1180 and also handles the present case, where the hash has no `=` at all but still gets wiped.

## Related

- #1180 — "Hash is removed if it contains an `=`." describes the same over-aggressive cleanup, but its title/description only covers the `=`-containing case. The actual condition is broader: the hash is cleared for **any** fragment, regardless of whether it looks OAuth-shaped.

## Workaround

Consumers can snapshot `window.location.hash` before `new RemoteStorage()` and restore it with `history.replaceState` afterwards (replaceState avoids firing a spurious `hashchange`).

## Version

`2.0.0-beta.8` — also present on `master` as of this writing.

## Impact

Any SPA that uses hash-based routing (`#/foo`) alongside remotestorage.js loses its route on every page load/refresh. The bug is especially confusing because it fires synchronously inside the constructor — the symptom looks like the app's own router is misbehaving, not the RS library.

silverbucket:
## Summary
- limit authorize hash cleanup to actual remoteStorage OAuth callback parameters
- preserve SPA and other application hashes that are unrelated to OAuth
- add Mocha and legacy Jaribu regression coverage for both callback and non-callback fragments

## Root cause
`Authorize._rs_init` cleared `location.hash` whenever `extractParams()` returned a truthy value. Since `extractParams()` always returns an object, non-OAuth fragments like `#/todos` and `#foo=bar` were being wiped during construction.

## Impact
Applications using hash-based routing no longer lose their current route when `RemoteStorage` is constructed, while real OAuth callback fragments are still cleaned up as before.

Closes https://github.com/remotestorage/remotestorage.js/issues/1373 and https://github.com/remotestorage/remotestorage.js/issues/1180

### DIFF
--- a/src/authorize.ts
+++ b/src/authorize.ts
@@ -47,6 +47,15 @@ export interface AuthorizeOptions {
 // This is set in _rs_init and needed for removal in _rs_cleanup
 let onFeaturesLoaded: EventHandler;
 
+function hasAuthCallbackParams (params: AuthResult): boolean {
+  return typeof params.access_token === 'string' ||
+    typeof params.code === 'string' ||
+    typeof params.error === 'string' ||
+    typeof params.remotestorage === 'string' ||
+    typeof params.state === 'string' ||
+    typeof params.rsDiscovery === 'object';
+}
+
 function extractParams (url?: string): AuthResult {
   // FF already decodes the URL fragment in document.location.hash, so use this instead:
   // eslint-disable-next-line
@@ -270,9 +279,10 @@ export class Authorize {
 
   static _rs_init = function (remoteStorage: RemoteStorage): void {
     const params = extractParams();
+    const hasCallbackParams = hasAuthCallbackParams(params);
     let location: Location;
 
-    if (params) {
+    if (hasCallbackParams) {
       location = Authorize.getLocation();
       location.hash = '';
     }
@@ -281,7 +291,7 @@ export class Authorize {
     onFeaturesLoaded = function(): void {
       let authParamsUsed = false;
 
-      if (!params) {
+      if (!hasCallbackParams) {
         remoteStorage.remote.stopWaitingForToken();
         return;
       }
--- a/src/authorize.ts
+++ b/src/authorize.ts
@@ -52,7 +52,6 @@ function hasAuthCallbackParams (params: AuthResult): boolean {
     typeof params.code === 'string' ||
     typeof params.error === 'string' ||
     typeof params.remotestorage === 'string' ||
-    typeof params.state === 'string' ||
     typeof params.rsDiscovery === 'object';
 }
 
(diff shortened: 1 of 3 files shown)


src/authorize.ts
```diff
@@ -47,6 +47,15 @@ export interface AuthorizeOptions {
 // This is set in _rs_init and needed for removal in _rs_cleanup
 let onFeaturesLoaded: EventHandler;
 
+function hasAuthCallbackParams (params: AuthResult): boolean {
+  return typeof params.access_token === 'string' ||
+    typeof params.code === 'string' ||
+    typeof params.error === 'string' ||
+    typeof params.remotestorage === 'string' ||
+    typeof params.state === 'string' ||
```
raucao: This would still clear params if only the `state` param is set, which is probably not what we want.

raucao (changes requested):
Funny, I just ran into the same issue, but it was only cosmetic in my case, so I didn't fix it immediately. Definitely an important bugfix!

raucao (approved):

PR #1378, opened 2026-04-26
Fix IndexedDB recovery for missing stores

### LINKED ISSUE #1376: IndexedDB v1→v2 upgrade can leave the DB without a 'nodes' store, then clean() hangs trying to recover it
## Summary

`IndexedDB.open` has two bugs that combine to leave RemoteStorage stuck during init when the on-disk DB ends up at version 1 with no object stores. In my app this surfaced as a startup hang with a fallback to LocalStorage; the exact wall-clock duration depends on browser internals (see "Net user-visible effect" below).

The setup is reproducible by any code that calls `indexedDB.open('remotestorage')` *without* a version (e.g. an app-side probe that wants to inspect the DB state before RS constructs). That call creates an empty v1 DB if none exists. RS then opens with `DB_VERSION = 2`, hits the upgrade path, and the bugs below kick in.

I'm filing this as one issue because the two bugs reinforce each other — fixing either alone leaves the user stuck.

## Bug 1 — upgrade-from-v1 skips creating `nodes`

[`indexeddb.ts:327-340`](https://github.com/remotestorage/remotestorage.js/blob/master/src/indexeddb.ts#L327-L340):

```ts
req.onupgradeneeded = function (event) {
  const db = req.result;
  log("[IndexedDB] Upgrade: from ", event.oldVersion, " to ", event.newVersion);

  if (event.oldVersion !== 1) {
    log("[IndexedDB] Creating object store: nodes");
    db.createObjectStore('nodes', {keyPath: 'path'});
  }

  log("[IndexedDB] Creating object store: changes");
  db.createObjectStore('changes', {keyPath: 'path'});
};
```

The guard `if (event.oldVersion !== 1)` assumes that any v1 DB already contains a `nodes` store. That's not enforced anywhere — a v1 DB can exist without `nodes` if e.g. a prior `indexedDB.open('remotestorage')` (no version) created the file as empty v1, or if a previous upgrade was aborted mid-flight. The result is a v2 DB with `changes` only, no `nodes`.

## Bug 2 — recovery path `clean()` doesn't close the open connection

[`indexeddb.ts:342-356`](https://github.com/remotestorage/remotestorage.js/blob/master/src/indexeddb.ts#L342-L356) detects the missing store and calls `IndexedDB.clean()`:

```ts
req.onsuccess = function () {
  clearTimeout(timer);

  // check if all object stores exist
  const db = req.result;
  if (!db.objectStoreNames.contains('nodes') || !db.objectStoreNames.contains('changes')) {
    log("[IndexedDB] Missing object store. Resetting the database.");
    IndexedDB.clean(name, function () {
      IndexedDB.open(name, callback);
    });
    return;
  }

  callback(null, req.result);
};
```

[`clean()` at lines 372-384](https://github.com/remotestorage/remotestorage.js/blob/master/src/indexeddb.ts#L372-L384):

```ts
static clean (databaseName: string, callback: () => void) {
  const req = indexedDB.deleteDatabase(databaseName);

  req.onsuccess = function () {
    log(`[IndexedDB] Deleted database "${databaseName}"`);
    callback();
  };

  // TODO check if this does anything as onabort does not exist on type according to ts
  req.onerror = (req as any).onabort = function (evt) {
    console.error('Failed to remove database "' + databaseName + '"', evt);
  };
}
```

Two problems here:

1. **The open connection is never closed.** The `onsuccess` closure still holds `req.result` (the open `IDBDatabase`) when it calls `clean()`. Per the IndexedDB spec, `deleteDatabase` will not proceed while any connection is open — it fires `onblocked` and waits for the connections to close. Since nothing closes the local `db`, deletion blocks indefinitely.
2. **`onblocked` is not handled.** Even if you wanted to surface the blocked state to the user, there's no `req.onblocked = ...` here — the request just hangs silently.

## Net user-visible effect

Tracing through the code: the original `IndexedDB.open` succeeds (`onsuccess` at line 342 fires), which clears the 10s timer at line 343. The missing-store branch then calls `clean()`, which never calls back. As a result, the recursive `IndexedDB.open(name, callback)` at line 350 **never starts** — and `_rs_init`'s Promise (line 394) never resolves *or* rejects. There's no per-feature timeout in `features.ts`, so the IDB feature can stay pending indefinitely.

In practice the app I observed this in did fall back to LocalStorage after a wall-clock delay in the ~10s range, but I haven't instrumented the source to confirm the exact mechanism — it could be browser-internal behavior around blocked `deleteDatabase`, GC of the abandoned IDBDatabase handle, or something else. The DB is never actually cleaned up: every reload reproduces the same hang until the user manually runs `indexedDB.deleteDatabase('remotestorage')` from DevTools.

The console will show `[IndexedDB] Missing object store. Resetting the database.` and then nothing — no `Deleted database` follow-up, because `clean()`'s `onsuccess` never fires.

## Reproduction

```js
// In a fresh tab
const r = indexedDB.open('remotestorage');
r.onsuccess = () => r.result.close();
// wait a tick, then load the RS-using app
```

The probe creates an empty v1 DB. RS opens at v2, the upgrade handler skips `createObjectStore('nodes')`, the missing-store check fires, `clean()` is called, and the IDB feature init hangs.

## Suggested fixes

1. Replace the version-based guard with an object-store existence check, so the upgrade handler creates whichever required store is missing regardless of the source version:

   ```ts
   req.onupgradeneeded = function (event) {
     const db = req.result;
     if (!db.objectStoreNames.contains('nodes')) {
       db.createObjectStore('nodes', {keyPath: 'path'});
     }
     if (!db.objectStoreNames.contains('changes')) {
       db.createObjectStore('changes', {keyPath: 'path'});
     }
   };
   ```

   The version-based guard incorrectly assumes that a v1 DB always has `nodes`. Schema-based gating handles the empty-v1 case while preserving the existing behavior for healthy v1 DBs.

2. Close the open connection before calling `clean()`, and add an `onblocked` handler to `clean()` so blocked deletions don't hang silently:

   ```ts
   if (!db.objectStoreNames.contains('nodes') || !db.objectStoreNames.contains('changes')) {
     db.close();
     IndexedDB.clean(name, function () { IndexedDB.open(name, callback); });
     return;
   }
   ```

   ```ts
   static clean (databaseName, callback) {
     const req = indexedDB.deleteDatabase(databaseName);
     req.onsuccess = () => callback(null);
     req.onerror = (evt) => callback(evt);
     req.onblocked = (evt) => {
       log('[IndexedDB] deleteDatabase blocked — another connection is still open');
       // could either fail-fast or wait for onsuccess
     };
   }
   ```

I'm happy to send a PR if either of those approaches would land — let me know if you'd prefer a different shape.

## Workaround

For now I'm working around this by detecting the corrupt schema before constructing RS:

1. Use `indexedDB.databases()` (Chrome 71+ / Safari 14+ / Firefox 126+) as an existence gate. If no `remotestorage` DB is present we do nothing — RS will create v2 fresh.
2. If the DB exists, open it without a version (no upgrade triggered) just long enough to read `objectStoreNames`, then close.
3. Only call `deleteDatabase` when a required store is missing (`nodes`, plus `changes` at v2). A healthy v1 DB with the `nodes` store is left alone for RS to upgrade normally — deleting it would discard cached nodes and any queued offline changes.

The existence gate matters: calling `indexedDB.open(name)` without a version when no DB exists creates an empty v1 — exactly the trap we're recovering from. It never does the `open()` until `databases()` has confirmed the DB is there.

silverbucket:
## Summary

- create missing IndexedDB object stores based on the actual schema during upgrades
- close the opened database connection before deleting a corrupt schema
- add coverage for empty v1 upgrade recovery and blocked cleanup visibility

Fixes #1376.

### DIFF
--- a/src/indexeddb.ts
+++ b/src/indexeddb.ts
@@ -329,14 +329,15 @@ class IndexedDB extends CachingLayer {
 
         log("[IndexedDB] Upgrade: from ", event.oldVersion, " to ", event.newVersion);
 
-        if (event.oldVersion !== 1) {
+        if (!db.objectStoreNames.contains('nodes')) {
           log("[IndexedDB] Creating object store: nodes");
           db.createObjectStore('nodes', {keyPath: 'path'});
         }
 
-        log("[IndexedDB] Creating object store: changes");
-
-        db.createObjectStore('changes', {keyPath: 'path'});
+        if (!db.objectStoreNames.contains('changes')) {
+          log("[IndexedDB] Creating object store: changes");
+          db.createObjectStore('changes', {keyPath: 'path'});
+        }
       };
 
       req.onsuccess = function () {
@@ -346,6 +347,7 @@ class IndexedDB extends CachingLayer {
         const db = req.result;
         if (!db.objectStoreNames.contains('nodes') || !db.objectStoreNames.contains('changes')) {
           log("[IndexedDB] Missing object store. Resetting the database.");
+          db.close();
           IndexedDB.clean(name, function () {
             IndexedDB.open(name, callback);
           });
@@ -371,6 +373,10 @@ class IndexedDB extends CachingLayer {
    */
   static clean (databaseName: string, callback: () => void) {
     const req = indexedDB.deleteDatabase(databaseName);
+
+    req.onblocked = function (evt) {
+      console.warn(`Deleting IndexedDB database "${databaseName}" is blocked by another open connection`, evt);
+    };
 
     req.onsuccess = function () {
       log(`[IndexedDB] Deleted database "${databaseName}"`);
--- a/test/unit/indexeddb.test.mjs
+++ b/test/unit/indexeddb.test.mjs
@@ -1,0 +1,88 @@
+import "mocha";
+import { expect } from "chai";
+import sinon from "sinon";
+
+import IndexedDB from "../../build/indexeddb.js";
+
+function createObjectStoreNames(stores) {
+  return {
+    contains(name) {
+      return stores.has(name);
+    }
+  };
+}
+
+describe("IndexedDB", () => {
+  let originalIndexedDB;
+
+  beforeEach(() => {
+    originalIndexedDB = globalThis.indexedDB;
+  });
+
+  afterEach(() => {
+    globalThis.indexedDB = originalIndexedDB;
+    sinon.restore();
+  });
+
+  describe(".open", () => {
+    it("creates missing object stores during upgrades regardless of old version", (done) => {
+      const stores = new Set();
+      const db = {
+        objectStoreNames: createObjectStoreNames(stores),
+        createObjectStore(name) {
+          stores.add(name);
+        }
+      };
+      const request = { result: db };
+
+      globalThis.indexedDB = {
+        open() {
+          return request;
+        }
+      };
+
+      IndexedDB.open("remotestorage", (err, openedDb) => {
+        expect(err).to.equal(null);
+        expect(openedDb).to.equal(db);
+        expect(stores.has("nodes")).to.equal(true);
+        expect(stores.has("changes")).to.equal(true);
+        done();
+      });
+
+      request.onupgradeneeded({
+        oldVersion: 1,
+        newVersion: 2
+      });
+      request.onsuccess();
+    });
+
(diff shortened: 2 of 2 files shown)


raucao (approved):
Solid! 👍

PR #1383, opened 2026-04-27
Stop emitting non-standard "; charset=binary" suffix

### LINKED ISSUE #1379: WireClient appends non-standard `; charset=binary` to binary uploads, breaks Chrome <img> rendering
## Summary

`WireClient.put` unconditionally appends `; charset=binary` to the `Content-Type` of any `ArrayBuffer` or `ArrayBufferView` body that doesn't already carry a charset:

https://github.com/remotestorage/remotestorage.js/blob/v2.0.0-beta.9/src/wireclient.ts (search for `charset=binary`)

```js
if ((!contentType.match(/charset=/)) && (body instanceof ArrayBuffer || isArrayBufferView(body))) {
  contentType += '; charset=binary';
}
```

This was originally added as a Firefox workaround in 2014. Twelve years on, the original quirk appears to be gone, the syntax is non-standard, and it now causes a different cross-browser rendering bug. I'd like to propose removing it.

## Historical context

- [#668](https://github.com/remotestorage/remotestorage.js/issues/668) (May 2014) reported that **Firefox 28 on Windows** was auto-injecting `; charset=UTF-8` into binary upload headers, producing things like `Content-Type: a; charset=utf-8; charset=UTF-8`. Other browsers (and Linux Firefox) were unaffected.
- [#744](https://github.com/remotestorage/remotestorage.js/pull/744) (the fix) added the `; charset=binary` prefix with the rationale: *"Firefox will always add a UTF-8 charset, unless charset=binary. Therefore, we stop setting charsets other than binary."*

### 1. The original Firefox quirk is gone

Firefox 28 → current is 12 years of networking-stack rewrites. I have not been able to reproduce the auto-`charset=UTF-8` behavior on modern Firefox (any platform). Anyone who can still reproduce it: please share — that would change the calculus.

### 2. `charset=binary` is non-standard

The [IANA character set registry](https://www.iana.org/assignments/character-sets/character-sets.xhtml) does not contain a `binary` charset. For binary media types (`image/jpeg`, `application/pdf`, `application/octet-stream`, etc.), the `charset` parameter is undefined per the IANA media-type registrations — `image/jpeg` lists "Optional parameters: N/A".

RFC 6657 ("Update to MIME regarding 'charset' Parameter Handling in Textual Media Types") explicitly scopes the `charset` parameter to textual types. Setting it on a binary type produces a syntactically-valid-but-semantically-meaningless header.

### 3. It actively breaks downstream consumers

Concrete repro from a remoteStorage app (`inbox-rs`, hosted on 5apps):

1. App stores a JPEG via `storeFile('image/jpeg', path, arrayBuffer)`.
2. WireClient PUTs with `Content-Type: image/jpeg; charset=binary`.
3. 5apps preserves the suffix on subsequent `GET` (correctly — it's just echoing what was sent).
4. App fetches the file and constructs `new Blob([buffer], { type: 'image/jpeg; charset=binary' })`.
5. **Chrome refuses to render an `` element whose `src` is a `URL.createObjectURL()` for that Blob.** Firefox and Safari render fine.

The workaround is to strip the `charset` parameter on the read side ([inbox-rs#105](https://github.com/silverbucket/inbox-rs/pull/105) discusses the workaround). 

### 4. The current code has no comment explaining why it's still there

Whoever opens `wireclient.ts` today gets no signal that this is load-bearing, vestigial, or contested. That's a separate but related code-health issue.

## Proposal

**Remove the `; charset=binary` suffix entirely.** For binary media types, omitting `charset` is the spec-correct behavior, and modern browsers handle it correctly without intervention.

If anyone *does* still hit a Firefox auto-charset issue on a current browser, please post a repro here — I'd want to see it before closing the door on the original bug. But absent that, I think we should let the workaround go.

## References

- Original report: [#668](https://github.com/remotestorage/remotestorage.js/issues/668)
- Original fix: [#744](https://github.com/remotestorage/remotestorage.js/pull/744)
- IANA character-set registry: https://www.iana.org/assignments/character-sets/character-sets.xhtml
- IANA `image/jpeg` registration: https://www.iana.org/assignments/media-types/image/jpeg
- RFC 6657: https://www.rfc-editor.org/rfc/rfc6657
- Downstream workaround: [silverbucket/inbox-rs#105](https://github.com/silverbucket/inbox-rs/pull/105)

silverbucket:
## Summary

Closes #1379. Supersedes #1380 (closed because the head branch was renamed to follow the project naming convention; no review comments to lose).

`WireClient.put` and `Dropbox.put` append `; charset=binary` to the `Content-Type` header for ArrayBuffer / binary uploads. The string `binary` is not a registered IANA character set (RFC 6657), and the suffix has no effect on transport — XHR/fetch sends the body bytes verbatim regardless of the `charset=` parameter.

The suffix is sticky: servers (5apps, Armadietto, anyone storing the request `Content-Type` verbatim) write it into stored metadata and echo it back on GET. Some browsers — Chrome among them — refuse to render an `` whose `Blob` type carries the suffix, producing a valid-looking `blob:` URL that never paints. Apps that pass `getFile()`'s `contentType` straight to a `Blob` constructor hit this.

## Approach

Two commits, two halves of the fix:

### 1. Stop emitting the suffix

| File | Change |
|---|---|
| `src/wireclient.ts` | Drop the suffix-append in `put()` (and the now-orphaned `isArrayBufferView` import). |
| `src/dropbox.ts` | Drop the matching suffix-append in `put()`. |
| `src/util.ts` | Keep the `charset=binary` branch in `shouldBeTreatedAsBinary` with a comment explaining it's there for backward compat with files uploaded by `<=2.0.0-beta.9`. The existing control-char heuristic handles new uploads. |
| `src/baseclient.ts` | Update example folder listings in the JSDoc to show clean `Content-Type` values. |
| `test/unit/wireclient.test.mjs` | Rename the two PUT charset tests to verify the new behavior (no suffix added; caller-supplied charset preserved). |

### 2. Strip the suffix on the read side too

Stops the suffix from leaking through `getFile().contentType` for files that already exist on servers with the legacy shape baked into their stored metadata. Without this half, apps would still see broken `` rendering for old files even after upgrading to the fixed library version.

| File | Change |
|---|---|
| `src/util.ts` | Add `stripLegacyCharsetBinary` helper. Removes only the `; charset=binary` parameter — legitimate ones like `charset=UTF-8` on `text/html` are preserved. |
| `src/wireclient.ts`, `src/dropbox.ts`, `src/googledrive.ts` | Apply the strip at the response boundary in each remote's GET handler. Internal binary-vs-text detection still uses the original (pre-strip) header so `shouldBeTreatedAsBinary` keeps firing correctly. |
| `test/unit/wireclient.test.mjs` | Update the existing assertion to expect the cleaned `contentType`; add a regression test that legitimate charsets are not stripped. |
| `test/unit/node-wireclient-suite.js` | Same assertion update for the legacy Jaribu suite. |

GET-side mock-server responses in tests still send the suffix (tolerance input) — the assertions verify the library sanitizes it before returning to the caller.

## Backward compatibility

- **Existing files on servers**: still served with `; charset=binary` by the storage backend, but the library now strips it before exposing `contentType` to callers.
- **Apps reading `getFile().contentType`**: previously got `; charset=binary` for legacy files; now get clean mime. If anyone wrote code that expects the suffix, that breaks — but the library already exposes `body` as `ArrayBuffer` vs string for the binary distinction, so the suffix as a signal would be redundant. Risk is low.
- **`shouldBeTreatedAsBinary`**: still public API, still recognizes the suffix when called with the raw header value. Only what's *returned* from `get()` is sanitized.
- **Cachinglayer suite tests**: put items locally with `same/type; charset=binary` and read them back. These exercise the local cache, not the WireClient response path — unaffected.

### DIFF
--- a/src/baseclient.ts
+++ b/src/baseclient.ts
@@ -300,12 +300,12 @@ export class BaseClient {
    *     "screenshot-20170902-1913.png": {
    *       "ETag": "6749fcb9eef3f9e46bb537ed020aeece",
    *       "Content-Length": 53698,
-   *       "Content-Type": "image/png;charset=binary"
+   *       "Content-Type": "image/png"
    *     },
    *     "screenshot-20170823-0142.png": {
    *       "ETag": "92ab84792ef3f9e46bb537edac9bc3a1",
    *       "Content-Length": 412401,
-   *       "Content-Type": "image/png;charset=binary"
+   *       "Content-Type": "image/png"
    *     }
    *   }
    * }
--- a/src/cachinglayer.ts
+++ b/src/cachinglayer.ts
@@ -13,7 +13,8 @@ import {
   globalContext,
   isDocument,
   isFolder,
-  pathsFromRoot
+  pathsFromRoot,
+  stripLegacyCharsetBinary
 } from './util';
 
 function getLatest (node: RSNode): RSItem {
@@ -155,7 +156,7 @@ abstract class CachingLayer {
             return {
               statusCode: 200,
               body: item.body || item.itemsMap,
-              contentType: item.contentType
+              contentType: stripLegacyCharsetBinary(item.contentType || null)
             };
           } else {
             return { statusCode: 404 };
@@ -178,7 +179,7 @@ abstract class CachingLayer {
             return {
               statusCode: 200,
               body: item.body || item.itemsMap,
-              contentType: item.contentType
+              contentType: stripLegacyCharsetBinary(item.contentType || null)
             };
           } else {
             return {statusCode: 404};
--- a/src/dropbox.ts
+++ b/src/dropbox.ts
@@ -7,6 +7,7 @@ import {
   applyMixins,
   isFolder,
   shouldBeTreatedAsBinary,
+  stripLegacyCharsetBinary,
   getJSONFromLocalStorage,
   getTextFromArrayBuffer,
   localStorageAvailable,
@@ -453,7 +454,7 @@ class Dropbox extends RemoteBase implements Remote {
         return {
           statusCode: status,
           body: body,
-          contentType: mime,
+          contentType: stripLegacyCharsetBinary(mime),
           revision: rev
         };
       });
--- a/src/dropbox.ts
+++ b/src/dropbox.ts
@@ -494,10 +494,6 @@ class Dropbox extends RemoteBase implements Remote {
       return {statusCode: 412, revision: savedRev};
     }
 
-    if ((!contentType.match(/charset=/)) && isBinaryData(body)) {
-      contentType += '; charset=binary';
-    }
-
     if (body.length > 150 * 1024 * 1024) {
       //https://www.dropbox.com/developers/core/docs#chunked-upload
       throw new Error("Cannot upload file larger than 150MB");
--- a/src/googledrive.ts
+++ b/src/googledrive.ts
@@ -5,6 +5,7 @@ import {
   isFolder,
   cleanPath,
   shouldBeTreatedAsBinary,
+  stripLegacyCharsetBinary,
   getJSONFromLocalStorage,
   getTextFromArrayBuffer,
   localStorageAvailable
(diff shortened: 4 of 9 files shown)


raucao:
@silverbucket Could you reopen the PR from this repo, please? I have added the same fix for cached documents to `CachingLayer#get()`, but I'm unable to push to your fork.

silverbucket:
@raucao https://github.com/remotestorage/remotestorage.js/pull/1390

PR #1385, opened 2026-04-27
Port some Jaribu suites to Mocha

silverbucket:
Chipping away at migrating from Jaribu -> Mocha, the smallest suites first, plus a reusable helper.

Closes nothing on its own — keeps #1270  alive.

## Summary

- **Ports four Jaribu suites to Mocha:**
  - `modules-suite` → folded into existing `remotestorage.test.mjs` (camelCase alias coverage for hyphenated module names).
  - `baseclient/types-suite` → `baseclient/types.test.mjs`.
  - `revisioncache-suite` → `revisioncache.test.mjs`.
  - `dropbox-suite` → two real coverage gaps merged into existing `dropbox.test.mjs`:
    - The `_initialFetchDone` guard at [`src/dropbox.ts:380`](https://github.com/remotestorage/remotestorage.js/blob/main/src/dropbox.ts#L380) for conditional GETs issued before the first `fetchDelta`.
    - Response handling when the server omits the `Content-Type` header.
- **New shared helper:** `test/helpers/backend-behavior.mjs` exports `backendBehavior(getFreshClient)`, a reusable Mocha equivalent of the legacy AMD `test/behavior/backend.js`. `dropbox.test.mjs` consumes it now; PR 3's `googledrive.test.mjs` will share the same helper. The AMD original stays in tree until `googledrive-suite.js` is gone (PR 3).
- **Glob fix:** `test/unit/*.test.mjs` → `test/unit/**/*.test.mjs` in `package.json` (mocha + lint scripts) so the new nested `baseclient/types.test.mjs` is picked up.

Net: 4 Jaribu suite files removed; CI still runs both runners.


silverbucket:
@raucao @DougReeder need a review here, just migrating some tests from jaribu -> mocha.

test/unit/dropbox-suite.js
```diff
@@ -1,1047 +0,0 @@
...
-              token: null
-            });
-            test.assertAnd(env.client.token, null);
-            test.assertAnd(env.client.userAddress, null);
-            test.done();
-          }
-        },
-
-        {
-          desc: "#configure sets 'connected' to true, once token is given",
```
DougReeder, line 313: I'm not seeing a corresponding test in dropbox.test.mjs; is the functionality tested here tested in some new tests, or just not something that needs a test?

test/unit/dropbox-suite.js
```diff
@@ -1,1047 +0,0 @@
...
-          run: function (env, test) {
-            env.client.configure({ token: 'foobarbaz' });
-            test.assert(env.client.connected, true);
-          }
-        },
-
-        {
-          desc: "#configure caches token and userAddress in localStorage",
```
DougReeder, line 321: Likewise, is this functionality still tested?

test/unit/dropbox-suite.js
```diff
@@ -1,1047 +0,0 @@
...
-                  "given_name": "John"
-                },
-                "email": "john.doe@example.com"
-              })
-            });
-            }, 10);
-          }
-        },
-
-        {
```
DougReeder, line 367: likewise, is there a new test for this functionality?

DougReeder (commented):
dropbox.delete() returns a file revision; should we be asserting it has the expected value?

raucao:
@silverbucket Did you see @DougReeder's review comments? It seems like you may have missed them...

silverbucket:
> @silverbucket Did you see @DougReeder's review comments? It seems like you may have missed them...

I saw them, i've just been a little backlogged, will get to this soon!

test/unit/dropbox-suite.js
silverbucket: Yes, the migrated configure tests verify that `userAddress`, `token`, and `refreshToken` are persisted, including the token-only network-error path.

test/unit/dropbox-suite.js
silverbucket: Yes, this is covered by the "fetches the user info when token but no `userAddress` is given" test, which verifies that connected becomes true after configuring a token.

test/unit/dropbox-suite.js
silverbucket: Good catch. I added an explicit test verifying that `get()` sends a `GET` request to the Dropbox download endpoint.

silverbucket:
> dropbox.delete() returns a file revision; should we be asserting it has the expected value?

Good catch. Adding the assertion revealed that `_deleteSimple()` was discarding the revision returned by Dropbox. It now includes that revision in the response, and both successful delete paths assert the expected value.

src/dropbox.ts
```diff
@@ -994,7 +995,10 @@ class Dropbox extends RemoteBase implements Remote {
         this.rs._emit('error', new Error(responseBody.error_summary));
       }
 
-      return Promise.resolve({statusCode: response.status});
+      return Promise.resolve({
+        statusCode: response.status,
+        revision: response.status === 200 ? responseBody.rev : undefined
```
DougReeder, line 1000: Does Dropbox not return status 201 when an item is created?

DougReeder (commented):
If I'm reading the tests correctly, it appears we're only testing using `fetch`, and not XHR.  Fetch is available in all modern envronments, so that's probably fine, but did we want to still test using XHR?  I don't know of mock for XHR.

test/unit/dropbox.test.mjs
```diff
@@ -703,6 +718,53 @@ describe('Dropbox backend', () => {
       expect(new Uint16Array(result.body)).to.deep.equal(data);
     });
 
+    it("triggers fetchDelta first when initial fetch isn't complete and ifNoneMatch is given", async () => {
```
DougReeder, line 721: This appears to be making different assertions about what happens than `#get fails if initial call to fetchDelta is not yet complete`.  I don't know the production code well.  Do the new assertions better reflect what we need the code to do?

src/dropbox.ts
silverbucket: This is for delete though

test/unit/dropbox.test.mjs
silverbucket: Yeah, I think the new test is more correct, though I could add another check for a 304 rather than just that a code is returned.

silverbucket:
> If I'm reading the tests correctly, it appears we're only testing using `fetch`, and not XHR. Fetch is available in all modern envronments, so that's probably fine, but did we want to still test using XHR? I don't know of mock for XHR.

I'll add some basic XHR coverage, but maybe we should consider dropping official support for XHR and just go with suggesting a polyfill like we do in the [nodejs.md](https://github.com/remotestorage/remotestorage.js/blob/master/docs/nodejs.md).

DougReeder (approved):
LGTM

PR #1386, opened 2026-04-27
Allow WebFinger discovery for localhost/private hosts by default

### LINKED ISSUE #1384: Discover() rejects localhost / private IPs after webfinger.js v3 upgrade (beta.9 regression)
## Summary

`rs.connect("user@localhost:8000")` (and any address whose host resolves to a localhost / private IP) fails immediately with `WebFingerError: private or internal addresses are not allowed`, before any HTTP request is made. This is a regression introduced by the webfinger.js v3 upgrade in #1370 (commit 490b5f7) and ships in `2.0.0-beta.9`.

## Reproduction

1. Run an RS server locally (e.g. Armadietto on `localhost:8000`).
2. In a browser app on `2.0.0-beta.9`, call `rs.connect("alice@localhost:8000")`.
3. Observe `error` event with `name: 'DiscoveryError'`. No outbound request fires.

## Root cause

webfinger.js v3 added a new `allow_private_addresses` security flag (defaults to `false`). When `false`, `WebFinger#resolveAndValidateHost` throws on any host matching its private/localhost regex *before* attempting the WebFinger lookup:

```js
// node_modules/webfinger.js/dist/webfinger.js
async resolveAndValidateHost(rawHost) {
  const normalized = WebFinger.normalizeHost(rawHost);
  if (!this.config.allow_private_addresses) {
    if (WebFinger.isPrivateAddress(normalized.host)) {
      throw new WebFingerError("private or internal addresses are not allowed");
    }
    await this.validateDNSResolution(normalized.hostname);
  }
  return normalized;
}
```

`src/discover.ts` constructs WebFinger without that flag:

```ts
const webFinger = new WebFinger({
  tls_only: false,
  uri_fallback: true,
  request_timeout: config.discoveryTimeout
});
```

…so localhost / private-IP storage is unreachable.

## Impact

- Local development against any RS server on `localhost`.
- Self-hosted users running their RS server on a private LAN address (`192.168.x.x`, `10.x.x.x`, etc.).
- E2E test suites that drive the real OAuth flow against a Dockerized RS server (which is how we hit it — see [silverbucket/inbox-rs#105](https://github.com/silverbucket/inbox-rs/pull/105)).

## Proposed fix

The protection makes sense as a default for *server-side* webfinger callers (the original SSRF concern in [silverbucket/webfinger.js#65](https://github.com/silverbucket/webfinger.js/issues/65)), but for a browser-side library that already runs under the same-origin / CORS sandbox it offers no extra security and only blocks legitimate localhost development.

Two reasonable options:

1. **Hard-code `allow_private_addresses: true` in `Discover`.** Browsers already gate cross-origin requests; the flag's threat model doesn't apply here. One-line patch.

2. **Surface it as a config option** on `RemoteStorage`, defaulting to `true` so beta.9's behavior is restored, and let server/Node embedders flip it off if they want.

silverbucket:
## Problem

`rs.connect("user@localhost:8000")` (and any address whose host resolves to a localhost / private IP) fails immediately with `WebFingerError: private or internal addresses are not allowed`, before any HTTP request is made.

This is a regression introduced by the webfinger.js v3 upgrade in #1370 (commit 490b5f7) and ships in `2.0.0-beta.9`. webfinger.js v3 added an `allow_private_addresses` config option that defaults to `false`, and `WebFinger#resolveAndValidateHost` now throws on any host matching its private/localhost regex *before* attempting the WebFinger lookup.

That breaks:

- Local development against any RS server on `localhost`.
- Self-hosted users running their RS server on a private LAN address (`192.168.x.x`, `10.x.x.x`, etc.).
- E2E test suites that drive the real OAuth flow against a Dockerized RS server (which is how we hit it — see [silverbucket/inbox-rs#105](https://github.com/silverbucket/inbox-rs/pull/105)).

## Fix

The SSRF protection makes sense for server-side webfinger callers (the original concern in [silverbucket/webfinger.js#65](https://github.com/silverbucket/webfinger.js/issues/65)), but remoteStorage.js runs in browsers, where cross-origin requests are already gated by the same-origin policy / CORS — the flag's threat model doesn't apply there and only blocks legitimate localhost / LAN development.

This PR takes option 2 from the issue: a new `discovery` config object, defaulting to `true` so beta.9's behaviour is restored, with non-browser embedders able to opt back into the SSRF guard.

```js
// Browser apps (default): localhost / LAN discovery works
const remoteStorage = new RemoteStorage();

// Non-browser embedder that wants the SSRF guard back
const remoteStorage = new RemoteStorage({
  discovery: {
    allowPrivateAddresses: false
  }
});
```

## Breaking changes

  - `discoveryTimeout` (a flat top-level constructor option added in beta.9) has
    been replaced by `discovery.timeout` inside the new `discovery` config object.
    Update any existing usage:

    ```js
    // before
    new RemoteStorage({ discoveryTimeout: 10000 });

    // after
    new RemoteStorage({ discovery: { timeout: 10000 } });

Closes #1384.

### DIFF
--- a/src/remotestorage.ts
+++ b/src/remotestorage.ts
@@ -160,7 +160,10 @@ enum ApiKeyType {
  *     conflict: true
  *   },
  *   cordovaRedirectUri: undefined,
- *   discoveryAllowPrivateAddresses: true,
+ *   discovery: {
+ *     allowPrivateAddresses: true,
+ *     timeout: 5000
+ *   },
  *   logging: false,
  *   modules: []
  * });
@@ -429,7 +432,17 @@ export class RemoteStorage {
   constructor (cfg?: object) {
     // Initial configuration property settings.
     // TODO use modern JS to merge object properties
-    if (typeof cfg === 'object') { extend(config, cfg); }
+    if (typeof cfg === 'object' && cfg !== null) {
+      // Pull out nested option groups so the shallow `extend` below doesn't
+      // overwrite the whole sub-object when the caller only sets a subset
+      // (e.g. `{ discovery: { timeout: 10 } }` keeps the default
+      // `allowPrivateAddresses`).
+      const { discovery, ...rest } = cfg as { discovery?: object };
+      extend(config, rest);
+      if (discovery && typeof discovery === 'object') {
+        extend(config.discovery, discovery);
+      }
+    }
 
     this.addEvents([
       'ready', 'authing', 'connecting', 'connected', 'disconnected',
--- a/src/remotestorage.ts
+++ b/src/remotestorage.ts
@@ -433,7 +433,7 @@ export class RemoteStorage {
     // Initial configuration property settings.
     // TODO use modern JS to merge object properties
     if (typeof cfg === 'object' && cfg !== null) {
-      // Pull out nested option groups so the shallow `extend` below doesn't
+      // Pull out nested option groups so the `Object.assign` below doesn't
       // overwrite the whole sub-object when the caller only sets a subset
       // (e.g. `{ discovery: { timeout: 10 } }` keeps the default
       // `allowPrivateAddresses`).
--- a/src/remotestorage.ts
+++ b/src/remotestorage.ts
@@ -160,6 +160,7 @@ enum ApiKeyType {
  *     conflict: true
  *   },
  *   cordovaRedirectUri: undefined,
+ *   discoveryAllowPrivateAddresses: true,
  *   logging: false,
  *   modules: []
  * });
@@ -173,6 +174,13 @@ enum ApiKeyType {
  * > [!TIP]
  * > For the change events configuration, you have to set all events
  * > explicitly.  Otherwise it disables the unspecified ones.
+ *
+ * > [!TIP]
+ * > `discoveryAllowPrivateAddresses` controls whether WebFinger discovery may
+ * > target localhost / private-IP storage servers. It defaults to `true`
+ * > because cross-origin requests in browsers are already gated by the
+ * > same-origin policy / CORS. Set it to `false` in non-browser embedders
+ * > that want webfinger.js's SSRF protection back.
  *
  * ## Events
  *
--- a/src/remotestorage.ts
+++ b/src/remotestorage.ts
@@ -175,13 +175,6 @@ enum ApiKeyType {
  * > For the change events configuration, you have to set all events
  * > explicitly.  Otherwise it disables the unspecified ones.
  *
- * > [!TIP]
- * > `discoveryAllowPrivateAddresses` controls whether WebFinger discovery may
- * > target localhost / private-IP storage servers. It defaults to `true`
- * > because cross-origin requests in browsers are already gated by the
- * > same-origin policy / CORS. Set it to `false` in non-browser embedders
- * > that want webfinger.js's SSRF protection back.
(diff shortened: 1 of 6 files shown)


src/remotestorage.ts
```diff
@@ -174,6 +175,13 @@ enum ApiKeyType {
  * > For the change events configuration, you have to set all events
  * > explicitly.  Otherwise it disables the unspecified ones.
  *
+ * > [!TIP]
+ * > `discoveryAllowPrivateAddresses` controls whether WebFinger discovery may
+ * > target localhost / private-IP storage servers. It defaults to `true`
+ * > because cross-origin requests in browsers are already gated by the
+ * > same-origin policy / CORS. Set it to `false` in non-browser embedders
+ * > that want webfinger.js's SSRF protection back.
+ *
```
raucao: This should not be a tip, if it's a normal config option. It should also not be documented in inline comments in the code, other than the correct TSDoc documentation comment(s).

raucao (changes requested):
The option key is a bit clunky, because it's so long. Wondering if we should turn that into an object instead and then change the timeout one, too (we can still do breaking changes for 2.0).

src/remotestorage.ts
```diff
@@ -428,7 +432,17 @@ export class RemoteStorage {
...
-    if (typeof cfg === 'object') { extend(config, cfg); }
+    if (typeof cfg === 'object' && cfg !== null) {
+      // Pull out nested option groups so the shallow `extend` below doesn't
+      // overwrite the whole sub-object when the caller only sets a subset
+      // (e.g. `{ discovery: { timeout: 10 } }` keeps the default
+      // `allowPrivateAddresses`).
+      const { discovery, ...rest } = cfg as { discovery?: object };
+      extend(config, rest);
+      if (discovery && typeof discovery === 'object') {
+        extend(config.discovery, discovery);
```
DougReeder:
```suggestion
      Object.assign(config, rest);
      if (discovery && typeof discovery === 'object') {
        Object.assign(config.discovery, discovery);
```

DougReeder (changes requested):
This is very good!

src/remotestorage.ts
```diff
@@ -433,7 +433,7 @@ export class RemoteStorage {
     // Initial configuration property settings.
     // TODO use modern JS to merge object properties
     if (typeof cfg === 'object' && cfg !== null) {
-      // Pull out nested option groups so the shallow `extend` below doesn't
+      // Pull out nested option groups so the `Object.assign` below doesn't
```
DougReeder, line 436: Ah, good catch!

DougReeder (approved):
LGTM

raucao (changes requested):
Looking great! 👏 

The PR description needs to mention the breaking change for the existing discovery timeout config key. At the moment, it only describes the fix.

silverbucket:
@raucao updated!

raucao:
`@param` docs can't be used like that for classes it seems:

Screenshot From 2026-05-27 12-33-27

I moved the documentation to a a more appropriate place, which fixes the issue and also makes it more prominent/relevant where it's needed: 71294118216324753072670b591170ce69a1b351

I also added a drive-by commit improving some wording on the same page.

raucao (approved):
👍

Issue #1388, opened 2026-04-29
Dropbox sharing


jonocodes:
I need to make some files publicly sharable via dropbox sharing. Which requires some additional scopes. Specifically 
_sharing.write_ and _sharing.read_.

But the oauth scopes for dropbox are hardcoded

https://github.com/remotestorage/remotestorage.js/blob/7e552dee922886869cee3001874504905dc1a65d/src/dropbox.ts#L56

I was able to patch my version such that these scopes were added and it worked for me.

I suggest we allow in config adding to the base scopes. If interested, I can work on it.

raucao:
Making it configurable seems reasonable. We could also add those to the default scopes.

Can you open a PR for it?

raucao:
By the way, there's also still an open PR for getting item URLs for Dropbox: #1130

jonocodes:
@raucao should this cover it?
https://github.com/remotestorage/remotestorage.js/pull/1389

raucao:
@jonocodes Thanks for the PR, and sorry for the slight delay! I'm going to have time for it sometime next week, if it still needs a review then.

raucao:
Closed in favor of #1052

PR #1389, opened 2026-05-04
Make getItemURL() async, implement for Dropbox

### LINKED ISSUE #1052: Problems with getItemURL on Dropbox

jonocodes:
## Summary

`BaseClient.getItemURL()` has never worked for the Dropbox backend — it was monkey-patched to throw `Error('getItemURL is not implemented for Dropbox yet')`. This PR implements it properly.

### What changed

- **`BaseClient.getItemURL`** is now `async`, returning `Promise<string | undefined>` instead of `string | undefined`. For standard remoteStorage backends the behavior is identical (resolves to `href + path`); for cloud backends it delegates to the remote's own implementation.
- **`Remote` interface** gains an optional `getItemURL?(path): Promise<string | undefined>` method — the TODO comment in `baseclient.ts` pointed at this for years.
- **`DropboxClient.getItemURL`** is implemented: checks the `_itemRefs` localStorage cache first, falls through to `share()` (which calls `POST /sharing/create_shared_link_with_settings`) on a miss.
- **`sharing.read` and `sharing.write` added to `OAUTH_SCOPE`** — these scopes are required for the Dropbox sharing API (`create_shared_link_with_settings` and `list_shared_links`) and were previously missing.
- **Monkey-patch removed from Dropbox** — `hookGetItemURL` / `unHookGetItemURL` in `dropbox.ts` are gone since Dropbox now implements a real `getItemURL`.
- **GoogleDrive** keeps `hookGetItemURL` / `unHookGetItemURL` — the error-throwing stub remains in place. Full implementation is left for future work.

### Breaking changes

1. **`getItemURL` is now async:** BaseClient.getItemURL() now returns Promise<string | undefined> instead of string | undefined. Callers must await the result:

    ```js
    // Before (only worked on standard remoteStorage backends)
    const url = client.getItemURL('public/photo.jpg');

    // After
    const url = await client.getItemURL('public/photo.jpg');
    ```

    Since the method previously threw for Dropbox and Google Drive, and the sync return value still works correctly for standard backends when awaited, the practical impact is limited to callers already using the standard backend.

2. **Dropbox OAuth scope changes:** sharing.read and sharing.write have been added to the hardcoded OAUTH_SCOPE for Dropbox to support generating shared links. Existing Dropbox users will need to re-authorize. 

    **Note for App Developers:** Developers using the Dropbox backend must update their app configuration in the Dropbox App Console (https://www.dropbox.com/developers/apps) and manually enable the sharing.read and sharing.write scopes on the "Permissions" tab. If the scopes are not enabled in the App Console, Dropbox will reject new authorization requests with an invalid_scope error until this is updated.

### Tests

- 6 new tests in `test/unit/dropbox.test.mjs` covering: cache hit, API call, `shared_link_already_exists` 409 fallback, non-public paths, folder paths, and unexpected API errors.
- 2 existing `baseclient-suite.js` tests updated to `await` the now-async method, with rejection handlers added (they were previously passing vacuously).
- 2 new `baseclient-suite.js` tests: `!connected → undefined` and delegation to `remote.getItemURL`.

---

## Notes

This is my first PR here, so let me know if I missed anything. 
Also I was not aware of https://github.com/remotestorage/remotestorage.js/pull/1130 until I finished making this. Sorry. However the approach is a bit different.

> Developed with AI assistance.

closes #1052
refs #1130, #1388


test/unit/baseclient-suite.js
```diff
@@ -424,8 +424,11 @@ define(['./build/config', './build/baseclient', 'test/helpers/mocks', 'tv4'],
...
           env.storage.remote = { href: 'http://example.com/test' };
 
-          var itemURL = env.client.getItemURL('A%2FB /C/%bla//D');
-          test.assert(itemURL, 'http://example.com/test/foo/A%252FB%20/C/%25bla/D');
+          env.client.getItemURL('A%2FB /C/%bla//D').then(function(itemURL) {
+            test.assert(itemURL, 'http://example.com/test/foo/A%252FB%20/C/%25bla/D');
+          }, function(err) {
+            test.result(false, err);
+          });
         }
```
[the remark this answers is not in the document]
jonocodes: @copilot apply changes based on [this feedback](https://github.com/remotestorage/remotestorage.js/pull/1389#discussion_r3179585880)

raucao (changes requested):
Thanks for the PR! The Dropbox implementation looks solid and the tests are great. :clap: 

However, we also have to change the Google Drive implementation for this: since `getItemURL()` is now async, the `hookGetItemURL` monkey patch in `src/googledrive.ts` breaks the promise contract by throwing a synchronous error instead of returning a rejected promise. This would crash apps that are expecting to use `.catch()` on it.

Instead of keeping the monkey patch, could you just drop `hookGetItemURL` and `unHookGetItemURL` completely and add this directly to the `GoogleDrive` class?

```ts
async getItemURL(_path: string): Promise<never> {
  throw new Error('getItemURL is not implemented for Google Drive yet');
}
```

That way, we can just delete the monkey-patching logic and properly satisfy the new async interface.

Also, the warning should stay in the function documentation, but updated for the new situation ("only works for remoteStorage and Dropbox"). We can remove it when it's finally implemented for Google Drive as well.

raucao:
FYI: I have updated the "Breaking change" section to include the fact that app developers offering Dropbox sync will also have to update their registered apps' permissions for this rs.js release/upgrade.

raucao:
We'll bundle this one together with #1386 in a single release, so that the breaking changes from both PRs are introduced at once.

@jonocodes Any chance you could have a look at my review comments sometime this coming week?

jonocodes:
@raucao Ok. I have cleaned up the hook handling for google drive.

src/baseclient.ts
```diff
@@ -639,25 +639,26 @@ export class BaseClient {
...
+   *          if no remote storage is connected. For standard remoteStorage
+   *          backends the URL is derived from the server's base href. For
+   *          Dropbox, a share link is fetched or created via the provider's API.
    *
-   * > [!WARNING]
-   * > This method currently only works for remoteStorage
-   * > backends. The GitHub issues for implementing it for Dropbox and Google
-   * > are 1052 and 1054.
+   * @warning Not yet implemented for Google Drive — throws an error when called
+   *          with that backend.
```
raucao: TSDoc does not have a `@warning` tag. We'll have to keep using the same syntax as before here.

raucao (changes requested):
Thanks!

Just one little problem left...

src/baseclient.ts
jonocodes: ok, updated

raucao: Thanks!

raucao:
FYI, I'm taking a bit longer trying to upgrade Sharesome for more than 10 years of changes in the JS and framework ecosystem, so I can properly test the integration with a real app (and modernize Sharesome in the process). I hope to have it done next week latest.

raucao (approved):
I have successfully tested the changes in my updated version of Sharesome. 👍 

However, I had to replace a URL param in order to load raw files from Dropbox, because public links are HTML pages by default:

https://github.com/raucao/sharesome/pull/87/changes/6d02f7a263b62871ebb1935b6bb53bfeeccfc939

If this is the behavior we want, then we can merge the PR. We'll still have to update [the documentation page](https://remotestorage.io/rs.js/docs/dropbox-and-google-drive.html) for the new permissions, so we could document this gotcha there as well (can be done in a separate PR).

raucao:
@jonocodes Can you confirm that getting the URL for the public sharing HTML page is the behavior you intended?

@remotestorage/core Anyone else has opinions on which URL we should get?

raucao:
*bump*

The above question is still open.

silverbucket:
@raucao You said you have to modify the URL to remove the params, in order to get the actual file and not an HTML page, but I'm not sure about your question "which URL we should get" ... we are getting one URL, and we may/may not need to modify it. Correct? At this point, the app developer needs to modify it if that's what they want. 

I see no point in doing it automatically, as we become locked into a hack around something in Dropbox which could change at any point. Maybe provide a simple utility Dropbox function to clean a given URL?

raucao:
It's not really "cleaning it". Have a look: https://github.com/raucao/sharesome/commit/6d02f7a263b62871ebb1935b6bb53bfeeccfc939

I agree that we should probably return whatever Dropbox returns, but a lot of apps will also want access to the raw document URL. Hence the question. I don't have a strong opinion on what to return, but I wouldn't want to introduce backend-specific extra functions.

I guess one reasonable solution would be to leave this decision (and the maintenance) to data modules.

silverbucket:
@raucao yeah so maybe not a helper function, but maybe provide it as an example function in the dropbox-related documentation.

raucao:
@jonocodes @silverbucket Thank you guys for the PR and review! I'll open a new one for the docs change.

PR #1390, opened 2026-06-02
Stop emitting non-standard "; charset=binary" suffix

### LINKED ISSUE #1379: WireClient appends non-standard `; charset=binary` to binary uploads, breaks Chrome <img> rendering

silverbucket:
Closes #1379. Supersedes #1380 and #1383 (reopened from main repo so collaborators can push directly to the branch).

`WireClient.put` and `Dropbox.put` append `; charset=binary` to the `Content-Type` header for ArrayBuffer / binary uploads. The string `binary` is not a registered IANA character set (RFC 6657), and the suffix has no effect on transport — XHR/fetch sends the body bytes verbatim regardless of the `charset=` parameter.

The suffix is sticky: servers (5apps, Armadietto, anyone storing the request `Content-Type` verbatim) write it into stored metadata and echo it back on GET. Some browsers — Chrome among them — refuse to render an `` whose `Blob` type carries the suffix, producing a valid-looking `blob:` URL that never paints. Apps that pass `getFile()`'s `contentType` straight to a `Blob` constructor hit this.

### 1. Stop emitting the suffix

| File | Change |
|---|---|
| `src/wireclient.ts` | Drop the suffix-append in `put()` (and the now-orphaned `isArrayBufferView` import). |
| `src/dropbox.ts` | Drop the matching suffix-append in `put()`. |
| `src/util.ts` | Keep the `charset=binary` branch in `shouldBeTreatedAsBinary` with a comment explaining it's there for backward compat with files uploaded by `<=2.0.0-beta.9`. The existing control-char heuristic handles new uploads. |
| `src/baseclient.ts` | Update example folder listings in the JSDoc to show clean `Content-Type` values. |
| `test/unit/wireclient.test.mjs` | Rename the two PUT charset tests to verify the new behavior (no suffix added; caller-supplied charset preserved). |

### 2. Strip the suffix on the read side too

Stops the suffix from leaking through `getFile().contentType` for files that already exist on servers with the legacy shape baked into their stored metadata.

| File | Change |
|---|---|
| `src/util.ts` | Add `stripLegacyCharsetBinary` helper. Removes only the `; charset=binary` parameter — legitimate ones like `charset=UTF-8` on `text/html` are preserved. |
| `src/wireclient.ts`, `src/dropbox.ts`, `src/googledrive.ts` | Apply the strip at the response boundary in each remote's GET handler. |
| `test/unit/wireclient.test.mjs` | Update the existing assertion to expect the cleaned `contentType`; add a regression test that legitimate charsets are not stripped. |
| `test/unit/node-wireclient-suite.js` | Same assertion update for the legacy Jaribu suite. |

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -213,6 +213,10 @@ export const getJSONFromLocalStorage = (key: string): { [key: string]: any } => 
  * @returns {boolean}
  */
 export const shouldBeTreatedAsBinary = (content: string | ArrayBuffer, mimeType: string): boolean => {
+  // The `charset=binary` branch is kept for backward compatibility: files uploaded
+  // with a release <=2.0.0-beta.9 carry that non-standard suffix in their stored
+  // Content-Type. New uploads no longer emit it; the control-char heuristic below
+  // handles them.
   // eslint-disable-next-line no-control-regex
   return !!((mimeType && mimeType.match(/charset=binary/)) || /[\x00-\x08\x0E-\x1F\uFFFD]/.test(content as string));
 };
--- a/src/util.ts
+++ b/src/util.ts
@@ -222,6 +222,27 @@ export const shouldBeTreatedAsBinary = (content: string | ArrayBuffer, mimeType:
 };
 
 /**
+ * Strip the legacy `; charset=binary` parameter from a Content-Type.
+ *
+ * Releases <=2.0.0-beta.9 appended this non-standard parameter to all binary
+ * PUTs, so files written back then still carry it in their stored Content-Type
+ * forever. We sanitize on read so callers never see it: some browsers refuse
+ * to render an `<img>` whose Blob type carries the suffix.
+ *
+ * Only the `charset=binary` parameter is removed; legitimate parameters like
+ * `charset=UTF-8` on `text/html` are preserved.
+ *
+ * @param {string} mimeType - The raw Content-Type as returned by the server.
+ * @returns {string} The Content-Type with any `; charset=binary` removed.
+ */
+export const stripLegacyCharsetBinary = (mimeType: string | null): string | null => {
+  if (!mimeType) {
+    return mimeType;
+  }
+  return mimeType.replace(/\s*;\s*charset=binary\b/i, '');
+};
+
+/**
  * Read data from an ArrayBuffer and return it as a string
  * @param {ArrayBuffer} arrayBuffer
  * @param {string} encoding
--- a/src/util.ts
+++ b/src/util.ts
@@ -218,7 +218,7 @@ export const shouldBeTreatedAsBinary = (content: string | ArrayBuffer, mimeType:
   // Content-Type. New uploads no longer emit it; the control-char heuristic below
   // handles them.
   // eslint-disable-next-line no-control-regex
-  return !!((mimeType && mimeType.match(/charset=binary/)) || /[\x00-\x08\x0E-\x1F\uFFFD]/.test(content as string));
+  return !!((mimeType && mimeType.match(/charset=binary/)) || /[\x00-\x08\x0E-\x1F\x80-\x9F\uFFFD]/.test(content as string));
 };
 
 /**
--- a/src/baseclient.ts
+++ b/src/baseclient.ts
@@ -300,12 +300,12 @@ export class BaseClient {
    *     "screenshot-20170902-1913.png": {
    *       "ETag": "6749fcb9eef3f9e46bb537ed020aeece",
    *       "Content-Length": 53698,
-   *       "Content-Type": "image/png;charset=binary"
+   *       "Content-Type": "image/png"
    *     },
    *     "screenshot-20170823-0142.png": {
    *       "ETag": "92ab84792ef3f9e46bb537edac9bc3a1",
    *       "Content-Length": 412401,
-   *       "Content-Type": "image/png;charset=binary"
+   *       "Content-Type": "image/png"
    *     }
    *   }
    * }
--- a/src/cachinglayer.ts
+++ b/src/cachinglayer.ts
@@ -13,7 +13,8 @@ import {
   globalContext,
   isDocument,
   isFolder,
-  pathsFromRoot
+  pathsFromRoot,
+  stripLegacyCharsetBinary
 } from './util';
 
 function getLatest (node: RSNode): RSItem {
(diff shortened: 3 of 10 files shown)


raucao (approved):

raucao:
Someone should have a look at my added commit. It's meant to fix the problem for documents that are already cached.

raucao:
> Someone should have a look at my added commit. It's meant to fix the problem for documents that are already cached.

@silverbucket ^^

DougReeder:
This would certainly improve things, but it's not really addressing the underlying issue: why does the library have to guess how to handle downloaded files?

silverbucket:
@DougReeder to answer your question, since RS returns text as strings and binary data as ArrayBuffer. We can't really rely on Content-Type alone, if it's not set/set incorrectly. So we need to do some guesswork to smartly handle things. Eliminating it would require a more thought out implementation and potentially a rewrite of the whole process.

silverbucket:
@DougReeder Sorry, I meant the remoteStorage.js client API, not the server. The HTTP response arrives as bytes. The library then decides whether getFile() should expose those bytes as an ArrayBuffer or decode them into a string.

Ideally, Content-Type would make that decision unambiguous, but it's unreliable. So we also need to inspect the response body as a fallback. That is the guess work.

charset=binary was effectively being used as an extra hint for that decision, but it is non-standard and leaks into the public contentType, where it causes browser issues.

src/util.ts
```diff
@@ -213,10 +213,35 @@ export const getJSONFromLocalStorage = (key: string): { [key: string]: any } =>
  * @returns {boolean}
  */
 export const shouldBeTreatedAsBinary = (content: string | ArrayBuffer, mimeType: string): boolean => {
+  // The `charset=binary` branch is kept for backward compatibility: files uploaded
+  // with a release <=2.0.0-beta.9 carry that non-standard suffix in their stored
+  // Content-Type. New uploads no longer emit it; the control-char heuristic below
+  // handles them.
   // eslint-disable-next-line no-control-regex
   return !!((mimeType && mimeType.match(/charset=binary/)) || /[\x00-\x08\x0E-\x1F\uFFFD]/.test(content as string));
```
DougReeder: We should probably include the C1 control codes as well: 0x80-0x9F

DougReeder (approved):

src/util.ts
silverbucket: done

PR #1392, opened 2026-06-17
Fix Dropbox disconnect

### LINKED ISSUE #1391: Dropbox: disconnect() fails with "Could not fetch user info." error
### Describe the Bug
When calling `disconnect()` on a `RemoteStorage` instance connected to a Dropbox account (using the PKCE flow with a refresh token), the library throws an unhandled error:
```
Error: Could not fetch user info.
    at Dropbox.configure (remotestorage.js:262)
```

### Steps to Reproduce
1. Connect a Dropbox storage account using the `remoteStorage` widget (leveraging the OAuth2 PKCE / refreshToken flow).
2. Click **Disconnect** in the widget to terminate the session.
3. Observe the console throwing an `Error: Could not fetch user info.` and triggering the global `rs.on('error', ...)` handler.

### Root Cause Analysis
In `src/remotestorage.ts`, the `disconnect()` method invokes `remote.configure()` to clear the active connection settings:
```typescript
  disconnect (): void {
    if (this.remote) {
      this.remote.configure({
        userAddress: null,
        href: null,
        storageApi: null,
        token: null,
        properties: null
      });
    }
    // ...
```
However, the `RemoteSettings` interface also defines `refreshToken` (used by Dropbox and Google Drive backends). Because `refreshToken` is omitted from the clearing payload, the Dropbox backend (`src/dropbox.ts`) still retains its `refreshToken` value, while `token` gets cleared to `null`.

Inside `Dropbox.configure()`:
```typescript
    if (this.refreshToken || this.token) {
      this.connected = true;
      if (this.userAddress) {
        this._emit('connected');
        writeSettingsToCache();
      } else {
        try {
          const info = await this.info(); // <--- Throws because this.token is null
          this.userAddress = info.email;
          this._emit('connected');
          writeSettingsToCache();
        } catch (err) {
          this.connected = false;
          this.rs._emit('error', new Error('Could not fetch user info.'));
          writeSettingsToCache.apply(this);
        }
      }
    }
```
Since `this.refreshToken` is still truthy, `configure()` sets `connected = true`. But since `this.userAddress` is now `null` (cleared by `disconnect`), it falls back to calling `this.info()` to fetch user profile details. This request fails because the access token is missing, resulting in the unhandled error event.

### Suggested Fix
Update `disconnect()` in `src/remotestorage.ts` to explicitly clear the `refreshToken` configuration field as well:
```typescript
  disconnect (): void {
    if (this.remote) {
      this.remote.configure({
        userAddress: null,
        href: null,
        storageApi: null,
        token: null,
        refreshToken: null,
        properties: null
      });
    }
```

raucao:
* Delete the refresh token, so that disconnect works properly
* Delete the cache for mapping filenames to public URLs

fixes #1391

### DIFF
--- a/src/dropbox.ts
+++ b/src/dropbox.ts
@@ -243,6 +243,7 @@ class Dropbox extends RemoteBase implements Remote {
       this.connected = false;
       if (hasLocalStorage) {
         localStorage.removeItem(SETTINGS_KEY);
+        localStorage.removeItem(`${SETTINGS_KEY}:shares`);
       }
       this.rs.setBackend(undefined);
     };
--- a/src/remotestorage.ts
+++ b/src/remotestorage.ts
@@ -728,6 +728,7 @@ export class RemoteStorage {
         href: null,
         storageApi: null,
         token: null,
+        refreshToken: null,
         properties: null
       });
     }
--- a/test/unit/dropbox.test.mjs
+++ b/test/unit/dropbox.test.mjs
@@ -304,6 +304,32 @@ describe('Dropbox backend', () => {
     });
   });
 
+  describe("disconnect", () => {
+    it("removes SETTINGS_KEY and the shares cache from localStorage", async () => {
+      const sharesKey = `${SETTINGS_KEY}:shares`;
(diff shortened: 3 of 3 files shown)


DougReeder (approved):
The test passes, and LGTM

PR #1393, opened 2026-06-22
Secure CI workflows against supply-chain attacks

DougReeder:
1.  CI: removes unneeded permissions
2. CI: pins version of 3rd-party actions
3. CI: drops Node.js v20, adds Node.js v26

To test: push branch to the master branch of your personal repo, in GitHub Actions tab, observe that test-and-lint and release-drafter complete successfully, and notify-website-repo only fails because "Parameter token or opts.auth is required"

### DIFF
--- a/.github/workflows/notify-website-repo.yml
+++ b/.github/workflows/notify-website-repo.yml
@@ -5,12 +5,14 @@ on:
     branches:
       - master
 
+permissions: {}
+
 jobs:
   notify:
     runs-on: ubuntu-latest
     steps:
       - name: Repository Dispatch
-        uses: peter-evans/repository-dispatch@v2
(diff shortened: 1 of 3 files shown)


raucao (approved):
👍 🙏

raucao:
@DougReeder Apparently, the actions were already locked down, and this PR made the runs fail on master (but for some reason not for the PR itself):

https://github.com/remotestorage/remotestorage.js/actions/runs/28010199071

DougReeder:
Good catch!  I've updated the Actions Permissions for the repository, trimming the version from the allowed actions, so they won't be blocked by policy.

If we want to be really secure, we could also pin our GitHub actions to a full-length commit SHA, and require SHAs for every action.

raucao:
OK, thanks.

It's definitely nicer to have these configs checked into the repo(s).

